PR #3, opened 2019-11-19
DEV-5767 Added the ability to use async to upsert transactions for mu…

MikeMcGarry:
…ltiple portfolios simultaneously while also batching each request to allow for more than 10,000 transactions to be upserted in a single call


tests/integration/cocoon/test_cocoon_transactions.py
```diff
@@ -166,6 +199,7 @@ def test_load_from_data_frame_transactions_success(
             identifier_mapping=identifier_mapping,
             property_columns=property_columns,
             properties_scope=properties_scope,
+            batch_size=batch_size
         )
 
         self.assertTrue(
```
rizwansaeed: Just looking at the assert, does this pick up failures correctly i.e. when the request fails do we get back an empty (valid) version which makes this pass or do we not get that back which means what we have is currently ok?

lusidtools/cocoon/cocoon.py
```diff
@@ -418,10 +418,10 @@ def load_portfolio_batch(
 
             sync_batches = [
```
rizwansaeed: Add a comment to explain what this is doing

tests/integration/cocoon/test_cocoon_transactions.py
MikeMcGarry: On a failure we get an exception back.

MikeMcGarry: However this only checks successes and you are correct that we need to check that the number of successes is greater than 0.

lusidtools/cocoon/cocoon.py
MikeMcGarry: Added

PR #4, opened 2019-11-21
DEV-5832: add correct_bond_price_quotes()

thomasdotred:
- add test for correct_bond_price_quotes()
- move command line app utility tests to unit test folder


lusidtools/cocoon/utilities.py
```diff
@@ -730,8 +730,34 @@ def parse_args(args: dict):
...
+
+    price_col = mapping[file_type]["quote_scalar"]["price"]
+    type_col = mapping[file_type]["quote_scalar"]["type"]
+    type_code = mapping[file_type]["quote_scalar"]["type_code"]
+    scale_factor = mapping[file_type]["quote_scalar"]["scale_factor"]
+
+    df["Prices Corrected"] = None
+
+    for index, row in df.iterrows():
+        if row[type_col] == type_code:
```
rizwansaeed: can make this a ternary assigment:

```
corrected_price = row[price_col] * scale_factor if row[type_col] == type_code else row[price_col]
```

tests/unit/cocoon/test_utilities.py
```diff
@@ -1249,3 +1249,297 @@ def test_checkargs_with_incorrect_type(self, _, function, param):
     def test_checkargs_with_invalid_argument(self, _, function, kwargs):
         with self.assertRaises(ValueError):
             function(**kwargs)
+
+    def test_scale_quote_of_type(self):
```
rizwansaeed: we probably need some more tests here:
- column contains missing/`None` quote values
- no `type_code`s match

lusidtools/cocoon/utilities.py
```diff
@@ -730,8 +730,34 @@ def parse_args(args: dict):
...
+            corrected_price = row[price_col] * scale_factor
+        else:
+            corrected_price = row[price_col]
+        df.at[
+            index, "Prices Corrected"] = corrected_price
```
rizwansaeed: This isn't really a "correction", perhaps rename to "__adjusted_quote"

If we're adding additional columns we might need to adopt a convention to avoid any potential column name clashes, I'd suggest prefixing with "__" (double underscore)

rizwansaeed: Actually, we probably need to check if our adjusted column already exists, we have 2 options:
1) raise a ValueError
2) pass in the name for the column and use that

tests/unit/cocoon/test_utilities.py
thomasdotred: - add test for None values for not_of type, and of_type
- add test for invalid type & price column names

lusidtools/cocoon/utilities.py
thomasdotred: - changed to `__adjusted_quote`
- ` df["__adjusted_quote"] = None` creates the few a few lines before. Is it really necessary to check it still exists?

lusidtools/cocoon/utilities.py
thomasdotred: changed

tests/unit/cocoon/test_utilities.py
thomasdotred: - added test for missing not_of_type values and missing of_type values
- added tests for invalid type and price column mappings

lusidtools/cocoon/utilities.py
thomasdotred: - changed to __adjusted_quote
- the `__adjusted_quote` column is created only a few lines before, why is it necessary to check this only a few lines later?

PR #10, opened 2019-11-26
DEV-5924 Updated Dates test and removed redundant test

MikeMcGarry:


.travis.yml
```diff
@@ -1,4 +1,4 @@
-language: minimal
+language: python
```
rizwansaeed: why can't you use `minimal`?

MikeMcGarry: Due to pip install in the deployment

PR #16, opened 2019-11-27
DEV-5940 Added tests to ensure that properties are always created. Mo…

MikeMcGarry:
…ved the property definition creation to use system instead of default for the data type

### DIFF
--- a/lusidtools/cocoon/utilities.py
+++ b/lusidtools/cocoon/utilities.py
@@ -19,8 +19,9 @@ from pathlib import Path
 import re
 from lusidtools.cocoon.dateorcutlabel import DateOrCutLabel
 import logging
-
+import time as default_time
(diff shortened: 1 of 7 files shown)


lusidtools/cocoon/utilities.py
```diff
@@ -940,3 +941,30 @@ def strip_whitespace(df: pd.DataFrame, columns: list) -> pd.DataFrame:
         )
 
     return stripped_df
+
+
+def create_scope_id(time_generator=None):
+    """
+    This function creates a unique ID based on the time since epoch for use
+    as a scope id.
+
```
thomasdotred: What is time_generator, should it be added to docs?

MikeMcGarry: Good find, yes I will add it!

PR #23, opened 2019-12-10
DEV-6000 Added in support for sub-holding-keys across portfolios and …

MikeMcGarry:
…holdings

# Pull Request Checklist

- [x] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/CONTRIBUTING.md)
- [x] Tests pass

# Description of the PR


.gitignore
```diff
@@ -247,4 +247,5 @@ ENV/
 env.bak/
 venv.bak/
 
-**/*secrets*
\ No newline at end of file
+**/*secrets*
+*.data/
```
rizwansaeed: if you make this `**/*.data` it will also look in all subfolders

lusidtools/cocoon/cocoon.py
```diff
@@ -328,6 +330,19 @@ def _convert_batch_to_models(
     # Get the data types of the columns to be added as properties
     property_dtypes = data_frame.loc[:, property_columns].dtypes
 
+    if kwargs["sub_holding_key_type"] == "key-values":
```
rizwansaeed: what is the `key-values` magic string? probably should be in the doc string

lusidtools/cocoon/cocoon.py
```diff
@@ -749,15 +785,37 @@ def load_from_data_frame(
...
-        api_factory=api_factory,
-        properties_scope=properties_scope,
-        file_type=file_type,
-        data_frame=data_frame,
-        property_columns=property_columns,
-        domain_lookup=domain_lookup,
-    )
+    open_api_types = getattr(lusid.models, domain_lookup[file_type]["top_level_model"]).openapi_types
+
+    if "sub_holding_keys" in open_api_types.keys() and "dict" in open_api_types["sub_holding_keys"]:
```
rizwansaeed: I think it would be useful to add a comment to say what this is doing

tests/integration/cocoon/test_cocoon_failures.py
```diff
@@ -239,6 +279,7 @@ def test_load_from_data_frame_failure(
         property_columns,
         properties_scope,
         file_type,
+        sub_holding_keys,
         expected_exception,
```
rizwansaeed: Raise a separate JIra to refactor this test to assert on the content of the error as well as the fact that one got raised, i.e. make sure the correct error gets raised

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -202,6 +208,142 @@ def setUpClass(cls) -> None:
...
+                    "tax_lots.settlement_date": None,
+                },
+                {
+                    "Isin": "ISIN Security Identifier",
+                    "Sedol": "SEDOL Security Identifier",
+                    "Currency": "is_cash_with_currency",
+                },
+                ["Prime Broker"],
+                "operations001",
+                ["Quantity"],
```
rizwansaeed: nice tests 👍

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -202,6 +208,142 @@ def setUpClass(cls) -> None:
...
                 lusid.models.Version,
             ],
```
rizwansaeed: Is there already a test that defines what happens if a column name is duplicated in the source file? (I'm presuming it just takes the first column it finds but we should check this works)

tests/unit/cocoon/test_properties.py
```diff
@@ -373,3 +373,41 @@ def test_create_property_values(
...
+        ],
+        [
+            "A full key",
+            ["Portfolio/Signal/Strategy"],
+            "Traders",
+            "Transaction",
+            ["Portfolio/Signal/Strategy"]
+        ],
+    ])
+    def test_infer_full_property_keys(self, _, partial_keys, properties_scope, domain, expected_outcome):
```
rizwansaeed: Just for completeness I would add `Transaction/Strategy` > `Transaction/Transaction/Strategy` to show that 2 part codes alway get interpreted as scope/code

.gitignore
MikeMcGarry: Okay great done.

lusidtools/cocoon/cocoon.py
MikeMcGarry: It is a flag which specifies if the sub-holding-keys should be a dictionary of property values (holdings) or just a list of property keys (portfolio). 

The way that we tell originally is by doing

open_api_types = getattr(lusid.models, domain_lookup[file_type]["top_level_model"]).openapi_types

    if "sub_holding_keys" in open_api_types.keys() and "dict" in open_api_types["sub_holding_keys"]:

We could just do this again here instead of passing through this magic keyword argument, what do you think?

lusidtools/cocoon/cocoon.py
MikeMcGarry: Added

tests/integration/cocoon/test_cocoon_failures.py
MikeMcGarry: Done.

tests/integration/cocoon/test_cocoon_holdings.py
MikeMcGarry: Thanks

tests/integration/cocoon/test_cocoon_holdings.py
MikeMcGarry: Yes this is correct. I have added two test cases. One where the first column is populated which works and the other where the first column is empty and it fails

tests/unit/cocoon/test_properties.py
MikeMcGarry: Done.

lusidtools/cocoon/cocoon.py
MikeMcGarry: I added some comments around the code for now. Haven't put anything in the DocString as I'm not sure on the best way to list kwargs in there.

rizwansaeed: I think the original version might be better as it prevents the user from having to know about inner workings and ergo reduces scope for errors and makes it easier to use

MikeMcGarry: Okay done

PR #27, opened 2019-12-19
Feature/dev 6200

slemasne:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR


lusidtools/pandas_utils/lusid_pandas.py
```diff
@@ -0,0 +1,41 @@
+import pandas as pd
+from flatten_json import flatten
+
+
+def lusid_respose_to_pd(lusid_response):
```
rizwansaeed: typo in "`repose`"

lusidtools/pandas_utils/lusid_pandas.py
```diff
@@ -0,0 +1,41 @@
...
+
+    """
+    This function takes a LUSID API response and attempts to convert the response into a Pandas DataFrame or Series.
+    Not all LUSID API responses are the same datatype.
+    Therefore we need to implement some if/else conditional logic to see how the response should be converted.
```
rizwansaeed: This comment is more an implementation detail rather than user documentation, I would move to inside the function

lusidtools/pandas_utils/lusid_pandas.py
```diff
@@ -0,0 +1,41 @@
+import pandas as pd
+from flatten_json import flatten
+
+
+def lusid_respose_to_pd(lusid_response):
+
+    """
+    This function takes a LUSID API response and attempts to convert the response into a Pandas DataFrame or Series.
+    Not all LUSID API responses are the same datatype.
```
rizwansaeed: List out the types of supported responses e.g. object containing `values`, dictionary etc.

lusidtools/pandas_utils/lusid_pandas.py
```diff
@@ -0,0 +1,41 @@
...
+    elif hasattr(lusid_response, "to_dict"):
+
+        # The API responses for single LUSID objects have a "to_dict" attribute
+        # Example "Get Instrument" returns a dictionary with some key-value pairs for one instrument
+
+        pandas_object = pd.Series(flatten(lusid_response.to_dict(), "."))
+
+    else:
+
+        raise TypeError(
```
rizwansaeed: I wonder if rather than throwing an exception we convert the whole response raw? If you are using `lusidtools` for general use (i.e. not necessarily using `load_from_data_frame`) you might want to get the response and pre-process in a DataFrame in which case this would be a useful function 

@MikeMcGarry @thomasdotred what do you think?

lusidtools/pandas_utils/lusid_pandas.py
rizwansaeed: should this be `lusid_respose_to_data_frame` rather than `pd`? (adopting whatever convention we've already used for DataFrame in method names)

rizwansaeed: I think we need to create the `__init__.py` and export the function

tests/unit/pandas_utils/test_lusid_pandas.py
```diff
@@ -0,0 +1,79 @@
+from pathlib import Path
+import lusid
+import lusid.models as models
+import unittest
+from lusidtools.pandas_utils.lusid_pandas import lusid_respose_to_pd
+import pandas as pd
+
+test_instrument_response = lusid.models.instrument.Instrument(
```
rizwansaeed: move these into the test class definition

tests/unit/pandas_utils/test_lusid_pandas.py
```diff
@@ -0,0 +1,79 @@
...
+
+        instrument_series = lusid_respose_to_pd(test_instrument_response)
+        self.assertIsInstance(instrument_series, pd.Series)
+
+    def test_transaction_alias_response_to_series(self):
+
+        transaction_alias_series = lusid_respose_to_pd(transaction_alias_response)
+        self.assertIsInstance(transaction_alias_series, pd.Series)
+
+    def test_response_to_pd_fail(self):
```
rizwansaeed: also add a test for passing in `None` (for completeness)

lusidtools/pandas_utils/lusid_pandas.py
slemasne: Thanks. That's corrected.

slemasne: Sure, that's been added in latest commit.

lusidtools/pandas_utils/lusid_pandas.py
MikeMcGarry: What do you mean by convert raw? The only exception here is if there is no to_dict method right? I think every model should have this method.

Also @slemasne I think this should always return a Pandas DataFrame. The Series and DataFrame have different attributes and methods (interfaces) so with this function you will never be sure what you will get back.

tests/unit/pandas_utils/test_lusid_pandas.py
MikeMcGarry: Can we please also check some of the values on the output as well?

lusidtools/pandas_utils/lusid_pandas.py
slemasne: Thanks. Corrected in latest commit.

lusidtools/pandas_utils/lusid_pandas.py
rizwansaeed: sorry I meant to convert any arbitrary object to a DataFrame

lusidtools/pandas_utils/lusid_pandas.py
slemasne: Thanks. Corrected in latest commit.

lusidtools/pandas_utils/lusid_pandas.py
slemasne: Thanks. Thats updated.

lusidtools/pandas_utils/lusid_pandas.py
slemasne: Agreed. Based on comments above, I wont return a pd.Series. Therefore **lusid_response_to_data_frame** is better name.

tests/unit/pandas_utils/test_lusid_pandas.py
slemasne: Thats done.

tests/unit/pandas_utils/test_lusid_pandas.py
slemasne: That's done.
New check for 
(1) None
(2) Some df values

PR #46, opened 2020-01-31
Feature/dev 6531 Add FIGI enrichment for instrument upserts

thomasdotred:
# Pull Request Checklist

- [ ] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/CONTRIBUTING.md)
- [ ] Tests pass

# Description of the PR

refactors the behaviour of enrich instruments to also enrich FIGIs.


lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1327,38 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
     d = val
 
     return d
+
+
+def enrich_existing_column(data_frame, mapping, field_name, enriched_column, constant_prefix):
+    # A column for name already exists and needs to be enriched
+    print(mapping)
```
MikeMcGarry: Please remove print statement

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1327,38 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
     d = val
 
     return d
+
+
+def enrich_existing_column(data_frame, mapping, field_name, enriched_column, constant_prefix):
+    # A column for name already exists and needs to be enriched
```
MikeMcGarry: This comment needs to be extended to include any provided field_name, previously it was only for instrument name explicitly.

lusidtools/cocoon/instruments.py
```diff
@@ -335,50 +347,29 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
...
     # Missing mapping for name altogether
     if "name" not in list(mapping_required.keys()):
         mapping_required["name"] = enriched_column_name
+    else:
+        enrich_existing_column(data_frame, mapping_required, "name", enriched_column_name, constant_prefix)
+
+    if "Figi" not in list(instrument_identifier_mapping.keys()):
+        instrument_identifier_mapping["Figi"] = enriched_column_figi
+    else:
+        enrich_existing_column(data_frame, instrument_identifier_mapping, "Figi", enriched_column_figi, constant_prefix)
```
MikeMcGarry: I believe someone could also provide Instrument/default/Figi in the mapping. You should call prepare_key (which is also in the instruments file) on the instrument identifier mapping keys first then check for the existence of the full key.

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1327,38 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+        mapping[field_name] = {"default": mapping[field_name][1:]}
+        mapping[field_name]["column"] = enriched_column
+
+    # Is a constant specified by the default nested dictionary specification
+    elif (
+        isinstance(mapping[field_name], dict)
+        and "default" in list(mapping[field_name].keys())
+        and "column" not in list(mapping[field_name].keys())
+    ):
+        mapping[field_name]["column"] = enriched_column
```
MikeMcGarry: Let's explicitly return the and mapping from here

lusidtools/cocoon/utilities.py
MikeMcGarry: Can you please add type hints, a docstring and add the @checkargs decorator now that you've moved this to a function?

lusidtools/cocoon/instruments.py
```diff
@@ -292,17 +292,28 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
         if identifier.is_unique_identifier_type
     ]
 
+def get_leaf(d, IDs):
+    for lusidID, dataID in d.items():
```
MikeMcGarry: Can you please add type annotations and a docstring to this function?

lusidtools/cocoon/instruments.py
```diff
@@ -335,50 +347,29 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
                 continue
             elif len(single_search[0].external_instruments) > 0:
                 name = single_search[0].external_instruments[0].name
+                figi = single_search[0].external_instruments[0].identifiers["Figi"].value
```
MikeMcGarry: Will there always be a Figi on the response?

lusidtools/cocoon/instruments.py
```diff
@@ -326,6 +337,7 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
     )
 
     names = []
+    figis = []
 
     for response in responses:
         name = np.NaN
```
MikeMcGarry: You may want to add figi = np.NaN here, otherwise you will use the figi from the previous succesful search in all subsequent searches until you find a new figi.

lusidtools/cocoon/instruments.py
```diff
@@ -292,17 +292,28 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
...
     ]
 
+def get_leaf(d, IDs):
+    for lusidID, dataID in d.items():
+        if isinstance(dataID, dict) and "column" not in dataID:
+            get_leaf(dataID, IDs)
+        elif isinstance(dataID, dict) and "column" in dataID:
+            IDs.append((lusidID, dataID["column"]))
+        elif dataID[0] != "$":
+            IDs.append((lusidID, dataID))
```
MikeMcGarry: If this is looking for a constant prefix, that prefix should be a function argument not hardcoded to be "$".  You can see that it is already passed into enrich_instruments as the argument constant_prefix.

lusidtools/cocoon/instruments.py
MikeMcGarry: You may find utilities.handle_nested_default_and_column_mapping interesting to look at (and potentially re-use) here.

tests/integration/cocoon/test_instruments.py
```diff
@@ -152,3 +177,4 @@ def test_enrich_instruments(
             ),
             self.loop,
         ).result()
+        print("Done")
```
MikeMcGarry: Please remove this print statement.

tests/integration/cocoon/test_instruments.py
```diff
@@ -117,32 +117,57 @@ def test_get_unique_identifiers(self, _, expected_outcome) -> None:
...
+            ],
+            [
+                "No figi column",
+                "data/global-fund-combined-instrument-master-no-figi.csv",
+                {"name": "instrument_name"},
+                {"Isin": "isin", "ClientInternal": "client_internal"},
+            ],
         ]
     )
+    @unittest.skip("Call to openfigi not optimised")
```
MikeMcGarry: I understand that the call to Openfigi is not optimised. However we can't be making a change to code and then commenting out the tests. This change may have broken the ability to enrich instrument names and we wouldn't know it.

lusidtools/cocoon/instruments.py
```diff
@@ -335,50 +347,29 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
...
+    enriched_column_figi = "LUSID.Figi.Enriched"
 
     data_frame[enriched_column_name] = names
+    data_frame[enriched_column_figi] = figis
 
     # Missing mapping for name altogether
     if "name" not in list(mapping_required.keys()):
         mapping_required["name"] = enriched_column_name
+    else:
+        enrich_existing_column(data_frame, mapping_required, "name", enriched_column_name, constant_prefix)
```
MikeMcGarry: As per comment in utilities, let's explicitly return the data_frame and mapping_required or instrument_identifier_mapping from these calls so as to not rely on them being modified in place by the function.

lusidtools/cocoon/utilities.py
thomasdotred: Done

lusidtools/cocoon/utilities.py
thomasdotred: # A column for provided field_name already exists and needs to be enriched

lusidtools/cocoon/instruments.py
thomasdotred: added constant_prefix as arg to get_leaf()

lusidtools/cocoon/instruments.py
thomasdotred: returns mapping as mapping_required

lusidtools/cocoon/utilities.py
thomasdotred: Added:
- type hints
- @checkargs
- numpy style docstring

tests/integration/cocoon/test_instruments.py
thomasdotred: removed the skip condition. I see that the way the test was originally written allows the test to pass provided no exceptions are thrown. In this way even if the rate limit is exceeded and openfigi starts throwing 500s, the test still passes as all the processing still executes properly.

lusidtools/cocoon/instruments.py
thomasdotred: Check for figi key [long or short] then use that as the column header to enrich

lusidtools/cocoon/instruments.py
```diff
@@ -328,59 +339,53 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
...
+    if id:
+        instrument_identifier_mapping = enrich_existing_column(data_frame, instrument_identifier_mapping, id,
+                                                  enriched_column_figi, constant_prefix)
+    else:
+        instrument_identifier_mapping["Figi"] = enriched_column_figi
+
+    # if "Figi" not in list(instrument_identifier_mapping.keys()):
+    #     instrument_identifier_mapping["Figi"] = enriched_column_figi
+    # else:
+    #     mapping_required = enrich_existing_column(data_frame, instrument_identifier_mapping, "Figi", enriched_column_figi, constant_prefix)
```
MikeMcGarry: Please delete

lusidtools/cocoon/instruments.py
```diff
@@ -292,6 +292,14 @@ def get_unique_identifiers(api_factory: lusid.utilities.ApiClientFactory):
...
     ]
 
+def get_leaf(d, IDs, constant_prefix):
+    for lusidID, dataID in d.items():
+        if isinstance(dataID, dict) and "column" not in dataID:
+            get_leaf(dataID, IDs, constant_prefix)
+        elif isinstance(dataID, dict) and "column" in dataID:
+            IDs.append((lusidID, dataID["column"]))
+        elif dataID[0] != constant_prefix:
+            IDs.append((lusidID, dataID))
```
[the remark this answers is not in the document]
MikeMcGarry: Get leaves

PR #61, opened 2020-02-22
import sample data

slemasne:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


lusidtools/general_utils/quick_func.py
```diff
@@ -0,0 +1,302 @@
+# This file contains a set of data manipulation functions
```
rizwansaeed: not a big fan of `quick_func.py` as a name, I'd generally prefer something more descriptive such as `data_loading_funcs.py`

lusidtools/general_utils/quick_func.py
```diff
@@ -0,0 +1,302 @@
...
+
+    client_portfolios = {}
+
+    for index, portfolio_group, portfolio in portfolios.itertuples():
+        client_portfolios.setdefault(portfolio_group, []).append(portfolio)
+
+    return client_portfolios
+
+
+def fetch_instrument_universe(csv_file, paper=False):
```
rizwansaeed: `paper=False` is a bit specific and horrible, is it possible parameterise the function instead?

slemasne:
> I'm assuming these functions are just a straight port of what was in the notebook repo and then also adding some tests (👍)?
> 
> It feels like most of these are reading a csv and mapping columns and could (should?) be replaced with the cocoon equivalents.
> 
> Do you have a feel for how much work that would be?

I did think the same as I was going through these functions. I think they were written before cocoon. I will replace what I can with cocoon. It will be a medium amount of work but I think its worth it for consistency.

PR #64, opened 2020-02-25
DEV-6860 add PG functionality

slemasne:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

This PR adds portfolio group functionality to cocoon. 

The new functionality should load a DF which:

(1) Has 1 or more portfolio groups
(2) Has 0 or more portfolios
(3) If portfolio group does not already exist, create group and add portfolios

There is an outstanding point to consider what we should do when a portfolio group already exists and the user is passing that existing portfolio group new portfolios. I see two options:

(1) We could use the update_portfolio_group method however that would require a new domain model in cocoon, as it uses the **_UpdatePortfolioGroupRequest_** model and not _**CreatePortfolioGroupRequest**_ .

(2) We could modify the code to delete a portfolio group if it already exists, then create a new portfolio group with the same code and add underlying portfolios. That would mean deleting current portfolio groups along with all their properties.


lusidtools/cocoon/cocoon.py
```diff
@@ -287,6 +287,47 @@ def load_instrument_property_batch(
 
         return results
 
+    @staticmethod
```
rizwansaeed: this may have been there before, but why do these need to be `@staticmethod`s?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+                    lusid.api.TransactionPortfoliosApi
+                ).create_portfolio(
+                    scope="test_scope_20200225",
+                    transaction_portfolio=create_portfolio_model(code),
+                )
+        except lusid.exceptions.ApiException as e:
+            if e.status == 404:
+                print(f"The portfolio {code} already exists")
+
+    @parameterized.expand(
```
rizwansaeed: Out off interest why is this paramertized if there is only 1 test case?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+            scope=scope,
+            data_frame=data_frame,
+            mapping_required=mapping_required,
+            mapping_optional=mapping_optional,
+            file_type="portfolio_group",
+            property_columns=property_columns,
+            properties_scope=properties_scope,
+        )
+
+        self.assertEqual(
```
rizwansaeed: We should add the failure error message to the assert when the test fails

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+                "data/portfolio_group_no_portfolio.csv",
+                {"code": "PortGroupCode", "display_name": "PortGroupDisplayName",},
+                {},
+                [],
+                None,
+                None,
+            ]
+        ]
+    )
+    def test_one_port_groups_no_portfolios(
```
rizwansaeed: as above, why `parameterized`?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+            scope=scope,
+            data_frame=original_data_frame,
+            mapping_required=mapping_required,
+            mapping_optional=mapping_optional,
+            file_type="portfolio_group",
+            property_columns=property_columns,
+            properties_scope=properties_scope,
+        )
+
+        self.assertEqual(
```
rizwansaeed: add failure to message

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+            first=responses["portfolio_groups"]["success"][0]._id,
+            second=lusid.models.ResourceId(
+                scope="port_group_load_test", code="PG_TEST_NOPG1"
+            ),
+        )
+
+    @parameterized.expand(
```
rizwansaeed: and again

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,239 @@
...
+from pathlib import Path
+import pandas as pd
+import lusid
+import lusid.models as models
+from lusidtools import cocoon as cocoon
+from parameterized import parameterized
+from lusidtools import logger
+
+
+class CocoonTestPortfolioGroup(unittest.TestCase):
```
rizwansaeed: I think we need some failure cases to ensure the unhappy path is easy for the user to deal with:

- portfolio not created when adding to group
- multiple instances of portfolio in source data

lusidtools/cocoon/cocoon.py
slemasne: I'm actually not sure. These were there before. @MikeMcGarry  - what do you think?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Good point. In next commit, I will pass arguments to the function as normal.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Good point. In next commit, I will pass arguments to the function as normal.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Good point. In next commit, I will pass arguments to the function as normal.

lusidtools/cocoon/cocoon.py
MikeMcGarry: They don't need to be static methods, but there is no reason for them to be class or instance methods.

lusidtools/cocoon/cocoon.py
```diff
@@ -287,6 +287,47 @@ def load_instrument_property_batch(
...
+    ) -> lusid.models.PortfolioGroup:
+        """
+        Upserts a batch of portfolios to LUSID
+        :param lusid.utilities.ApiClientFactory api_factory: The api factory to use
+        :param list[lusid.models.CreateTransactionPortfolioRequest] portfolio_batch: The batch of portfolios to create
+        :param kwargs: 'scope', 'code' arguments required for the API call
+        :return: lusid.models.PortfolioGroup: The response from LUSID
+        """
+
+        updated_request = group_request_into_one("CreatePortfolioGroupRequest", portfolio_group_batch, ["values"])
```
MikeMcGarry: portfolio_group_batch[0].__class__.__name__ will get you "CreatePortfolioGroupRequest" and will also work if the name of this class ever changes.

lusidtools/cocoon/config/domain_settings.json
```diff
@@ -59,5 +59,16 @@
       "full_key_format": false,
       "required_call_attributes": [],
       "unique_attributes": []
+  },
+
+    "portfolio_group": {
+      "domain": "portfoliogroups",
+      "batch_allowed": true,
+      "default_batch_size": 2000,
```
MikeMcGarry: Is there a limit on the number of portfolios you can add in the creation request?

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+    Description
+    ------------
+    This function take a list of requests and collates an attribute from each request, adding the collated attributes
+    back onto the first request in the list. The function returns the modified first request.
+    For example, the function can take a list of CreatePortfolioGroupRequests, extract the "values" or portfolios from
+    each request, and then add all portfolios back onto the first request in the list.
+    :param lusid.models model_type:
+    :param list request_list:
+    :param list attribute_for_grouping:
+    :return: a single LUSID request
```
MikeMcGarry: Please add descriptions here

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+    """
+
+    # Define a base request for modifying - this is the first request in the list by default
+
+    base_request = request_list[batch_index]
+
+    for attrib in attribute_for_grouping:
+
+        if getattr(models, model_type).openapi_types[attrib] == "list[ResourceId]":
```
MikeMcGarry: I'd update this to just check if "list[" in getattr(models, model_type).openapi_types[attrib] so that this will work for lists of things other than ResourceIds

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+
+            # Collect the attributes from each request onto a list
+
+            batch_attrib = [
+                attrib
+                for nested_list in [
+                    attrib.values
+                    for attrib in request_list
+                    if attrib.values is not None
```
MikeMcGarry: You have hardcoded "values" in here. It also looks like you have used attrib in "for attrib in attribute_for_grouping" as well as "for attrib in request_list" when these refer to different things. 

Consider renaming "for attrib in request_list" to be "for request in request_list" and then use getattr(request, attrib) rather than attrib.values.

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+                ]
+                for attrib in nested_list
+            ]
+
+            # Assign collated values onto the base request
+
+            setattr(base_request, "values", batch_attrib)
```
MikeMcGarry: Values hardcoded here too

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+
+        elif (
+            getattr(models, model_type).openapi_types[attrib]
+            == "dict(str, ModelProperty)"
+        ):
```
MikeMcGarry: Same as above make this work with "dict(" so you can support other models.

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+                [
+                    (attrib, nested_list[attrib])
+                    for nested_list in [
+                        attrib.properties
+                        for attrib in request_list
+                        if attrib.properties is not None
+                    ]
+                    for attrib in nested_list
+                ]
+            )
```
MikeMcGarry: Same as above regarding re-use of attrib

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,72 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+
+            # Assign collated values onto the base request
+
+            setattr(base_request, "properties", batch_attrib)
```
MikeMcGarry: Properties hardcoded

lusidtools/cocoon/cocoon.py
slemasne: Thanks

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: sure, thats done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: sure, done.

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,52 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
 
         with self.assertRaises(expected_exception):
             cocoon.utilities.get_swagger_dict(api_url=api_url)
+
+
+    def test_group_request_into_one_list(self):
```
MikeMcGarry: Nice add in a test for something other than a CreatePortfolioGroupRequest

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,70 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+    base_request = request_list[batch_index]
+
+    for attrib in attribute_for_grouping:
+
+        if "list" in getattr(models, model_type).openapi_types[attrib]:
+
+            # Collect the attributes from each request onto a list
+
+            batch_attrib = [
+                attrib
```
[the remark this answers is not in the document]
slemasne: As discussed, names updated.

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,70 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+                for nested_list in [
+                    getattr(response, attrib)
+                    for response in request_list
+                    if getattr(response, attrib) is not None
+                ]
+                for attrib in nested_list
```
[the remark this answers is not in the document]
slemasne: As discussed, names updated.

lusidtools/cocoon/cocoon.py
```diff
@@ -287,6 +294,80 @@ def load_instrument_property_batch(
...
+                if code not in current_portfolio_group.portfolios
+            ]
+
+            for code, scope in set(
+                [(resource.code, resource.scope) for resource in new_portfolios]
+            ):
+                api_factory.build(lusid.api.PortfolioGroupsApi).add_portfolio_to_group(
+                    scope=kwargs["scope"],
+                    code=kwargs["code"],
+                    effective_at=datetime.now(tz=pytz.UTC),
```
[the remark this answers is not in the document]
slemasne: As discussed, this is required parameter.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                    port_group
+                    for nested_group in [
+                        port_group.portfolios
+                        for port_group in responses["portfolio_groups"]["success"]
+                    ]
+                    for port_group in nested_group
+                ]
+            ),
+            len(data_frame),
+        )
```
MikeMcGarry: Nice assertion!

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                [
+                    port_group._id
+                    for port_group in responses["portfolio_groups"]["success"]
+                ]
+            ),
+            second=len(data_frame),
+        )
+
+        self.assertEqual(
+            first=responses["portfolio_groups"]["success"][0]._id,
```
MikeMcGarry: Why do you have an underscore in front of _id? You should pick of the id attribute with no underscore.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+            second=lusid.models.ResourceId(scope=self.scope, code="PG_TEST3A"),
```
MikeMcGarry: Should this code come from the dataframe after importing the file in case you change the file?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                    port_group._id
+                    for port_group in responses["portfolio_groups"]["success"]
+                ]
+            ),
+            second=len(data_frame),
+        )
+
+        self.assertEqual(
+            first=responses["portfolio_groups"]["success"][1]._id,
+            second=lusid.models.ResourceId(scope=self.scope, code="PG_TEST5B"),
```
MikeMcGarry: Same comments as previous test about _id and the code being hard coded.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+            },
+            mapping_optional={},
+            file_type="portfolio_group",
+            property_columns=[],
+            properties_scope=None,
+        )
+
+        if responses.get("portfolio_groups").get("errors") != []:
+            for error in responses["portfolio_groups"]["errors"]:
+                print(json.loads(error.body)["title"])
```
MikeMcGarry: Can you log this instead of printing it please?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
```
MikeMcGarry: Can you please log this instead of printing it?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+            },
+            mapping_optional={"values.scope": "Scope", "values.code": "FundCode",},
+            file_type="portfolio_group",
+            property_columns=[],
+            properties_scope=None,
+        )
+
+        if responses.get("portfolio_groups").get("errors") != []:
+            for error in responses["portfolio_groups"]["errors"]:
+                print(json.loads(error.body)["title"])
```
MikeMcGarry: Please log instead of printing

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Better practice here would be to check that the length of the list is greater than 0.

Why did you use .get here "responses.get("portfolio_groups").get("errors")" and then [] syntax a few lines later "responses["portfolio_groups"]["errors"]"

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Better practice here would be to check that the length of the list is greater than 0.

Why did you use .get here "responses.get("portfolio_groups").get("errors")" and then [] syntax a few lines later "responses["portfolio_groups"]["errors"]"

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Better practice here would be to check that the length of the list is greater than 0.

Why did you use .get here "responses.get("portfolio_groups").get("errors")" and then [] syntax a few lines later "responses["portfolio_groups"]["errors"]"

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+
+        self.assertEqual(
+            json.loads(responses["portfolio_groups"]["errors"][0].body)["name"],
+            "PortfolioNotFound",
+        )
```
MikeMcGarry: Can we also assert that there are no successes and only failures in this case please?

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+            },
+            mapping_optional={"values.scope": "Scope", "values.code": "FundCode",},
+            file_type="portfolio_group",
+            property_columns=[],
+            properties_scope=None,
+        )
+
+        if responses.get("portfolio_groups").get("errors") != []:
+            for error in responses["portfolio_groups"]["errors"]:
+                print(json.loads(error.body)["title"])
```
MikeMcGarry: Better practice here would be to check that the length of the list is greater than 0.

Why did you use .get here "responses.get("portfolio_groups").get("errors")" and then [] syntax a few lines later "responses["portfolio_groups"]["errors"]"

MikeMcGarry: Please log instead of printing

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                    port_group
+                    for nested_group in [
+                        port_group.portfolios
+                        for port_group in responses["portfolio_groups"]["success"]
+                    ]
+                    for port_group in nested_group
+                ]
+            ),
+            len(duplicate_df["unique_combinations_for_test"].unique()),
+        )
```
MikeMcGarry: Am I right in saying that if you provide a duplicate in your request LUSID handles this gracefully and doesn't throw an error? I didn't see anything in our code to eliminate duplicates in the event of a creation request unlike where you take the set during an update.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                "display_name": "PortGroupDisplayName",
+            },
+            mapping_optional={"values.scope": "Scope", "values.code": "FundCode",},
+            file_type="portfolio_group",
+            property_columns=[],
+            properties_scope=None,
+        )
+
+        if responses.get("portfolio_groups").get("errors") != []:
+            for error in responses["portfolio_groups"]["errors"]:
```
MikeMcGarry: Better practice here would be to check that the length of the list is greater than 0.

Why did you use .get here "responses.get("portfolio_groups").get("errors")" and then [] syntax a few lines later "responses["portfolio_groups"]["errors"]"

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                print(json.loads(error.body)["title"])
```
MikeMcGarry: Please log instead of printing

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                scope=self.scope, code=first_group_series["PortGroupCode"]
+            )
+            .portfolios
+        )
+
+        self.assertIn(first_group_resoruce_id, get_first_group_response_portfolios)
+        self.assertNotIn(
+            lusid.models.ResourceId(scope="test_scope_20200225", code="TEST_COM5"),
+            get_first_group_response_portfolios,
+        )
```
MikeMcGarry: What is the significance of checking that this portfolio does not exist in the group?

MikeMcGarry: You are missing a test for updating a portfolio group when a portfolio that you are updating with does not exist. Please document the behaviour for this with a test.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+import pandas as pd
+import json
+import lusid
+import lusid.models as models
+from lusidtools import cocoon as cocoon
+from lusidtools.cocoon.utilities import create_scope_id
+from lusidtools import logger
+
+
+class CocoonTestPortfolioGroup(unittest.TestCase):
```
MikeMcGarry: Nice test coverage. Note that there is quite a lot of repeated code in the tests below. Consider using paramaterized or creating functions for some of the commonly used elements like you did with create_portfolio_model.

MikeMcGarry: Some gaps I did notice:

- All your test cases have portfolios in the same scope
- There is no test case for trying to update a group with a duplicate portfolio
- There is no test for trying to update a group when one of the portfolios you are trying to add does not exist
- There is no test case for trying to update a group when there are no new portfolios to add (i.e. all portfolios already added)
- There is no test case for creating a portfolio group with properties

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,69 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+                        for request in request_list
+                        if getattr(request, attrib) is not None
+                    ]
+                    for lusid_model in nested_list
+                ]
+            )
+
+            # Assign collated values onto the base request
+
+            setattr(base_request, attrib, batch_attrib)
```
MikeMcGarry: Very nice!! :) Great work 👍

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,69 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
     d = val
 
     return d
+
+
+def group_request_into_one(
+    model_type, request_list: list, attribute_for_grouping: list, batch_index=0
```
MikeMcGarry: Add type annotation model_type: str

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,111 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
 
         with self.assertRaises(expected_exception):
             cocoon.utilities.get_swagger_dict(api_url=api_url)
+
```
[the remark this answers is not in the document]
MikeMcGarry: There is a missing test case for:

- Providing a string of a for a model which doesn't exist e.g. "NotALusidModel"
- Providing an empty list
- Providing an index greater than the length of the list

lusidtools/cocoon/utilities.py
```diff
@@ -1327,3 +1328,69 @@ def update_value(d: typing.Union[dict, str], val: typing.Union[str, float]):
...
+):
+
+    """
+    Description
+    ------------
+    This function take a list of requests and collates an attribute from each request, adding the collated attributes
+    back onto the first request in the list. The function returns the modified first request.
+    For example, the function can take a list of CreatePortfolioGroupRequests, extract the "values" or portfolios from
+    each request, and then add all portfolios back onto the first request in the list.
+    :param lusid.models model_type: the model type which we will modify (eg CreatePortfolioGroupRequest).
```
MikeMcGarry: This is actually a string not a lusid.models class

lusidtools/cocoon/config/domain_settings.json
```diff
@@ -59,5 +59,16 @@
...
+      "top_level_model": "CreatePortfolioGroupRequest",
+      "portfolio_specific": true,
+      "full_key_format": false,
+      "required_call_attributes": ["scope"],
```
MikeMcGarry: code and effective_at are missing from the required_call_attributes list in the config here.

lusidtools/cocoon/cocoon.py
```diff
@@ -287,6 +294,80 @@ def load_instrument_property_batch(
...
+        if "scope" not in list(kwargs.keys()):
+            raise KeyError(
+                "You are trying to load a portfolio group without a scope, please ensure that a scope is provided."
+            )
+
+        if "code" not in list(kwargs.keys()):
+            raise KeyError(
+                "You are trying to load a portfolio group without a portfolio code, please ensure that a code is provided."
+            )
+
```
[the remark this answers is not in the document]
slemasne: As discussed, we will hard code to "now" as this is only required for the add_portfolio_to_group method.

lusidtools/cocoon/cocoon.py
```diff
@@ -287,6 +294,80 @@ def load_instrument_property_batch(
...
+                    code=kwargs["code"],
+                    effective_at=datetime.now(tz=pytz.UTC),
+                    portfolio_id=lusid.models.ResourceId(scope=scope, code=code),
+                )
+
+            print(
+                api_factory.build(lusid.api.PortfolioGroupsApi).get_portfolio_group(
+                    scope=kwargs["scope"], code=kwargs["code"]
+                )
+            )
```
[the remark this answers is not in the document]
slemasne: Removed from latest commit.

lusidtools/cocoon/cocoon.py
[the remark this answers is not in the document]
slemasne: I have added error handling to the loop of updates.
The desired behaviour - the function should update the portfolio group with portfolios if possible. One bad **add_portfolio_to_group** request should not preclude the remainder of the update.

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
slemasne: Moved test from integration to unit.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+
+        responses = cocoon.cocoon.load_from_data_frame(
+            api_factory=self.api_factory,
+            scope=self.scope,
+            data_frame=data_frame,
+            mapping_required={
+                "code": "PortGroupCode",
+                "display_name": "PortGroupDisplayName",
+            },
+            mapping_optional={"values.scope": "Scope", "values.code": "FundCode",},
```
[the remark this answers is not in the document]
slemasne: Thanks. I copied that from one of Toby's tests!

lusidtools/cocoon/cocoon.py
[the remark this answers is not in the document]
slemasne: This should be removed in latest commit.

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,111 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
...
+            ),
+        ]
+
+        # Run list tests
+
+        list_grouped_request = group_request_into_one(
+            "CreatePortfolioGroupRequest", requests, ["values"]
+        )
+        self.assertEqual(len(list_grouped_request.values), 4)
+        self.assertEqual(list_grouped_request.values[3].code, "PORT4")
```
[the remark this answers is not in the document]
slemasne: Sure, done.

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,111 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
...
+
+        #  Run dict tests
+
+        dict_grouped_request = group_request_into_one(
+            "CreatePortfolioGroupRequest", requests, ["properties"]
+        )
+        self.assertEqual(len(dict_grouped_request.properties), 3)
+        self.assertEqual(
+            list(dict_grouped_request.properties.values())[0].value, "prop1"
+        )
```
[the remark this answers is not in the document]
slemasne: sure, done.

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,111 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
...
+            ),
+        ]
+
+        # Run list tests
+
+        list_grouped_request = group_request_into_one(
+            "HoldingAdjustment", holding_requests, ["tax_lots"]
+        )
+        self.assertEqual(len(list_grouped_request.tax_lots), 2)
+        self.assertEqual(list_grouped_request.tax_lots[1].purchase_date, "2020-02-20")
```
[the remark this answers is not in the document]
slemasne: Sure, done.

tests/integration/cocoon/test_utilities.py
```diff
@@ -28,3 +31,111 @@ def test_get_swagger_dict_fail(self, _, api_url, expected_exception):
...
+        #  Run dict tests
+
+        dict_grouped_request = group_request_into_one(
+            "CreatePortfolioGroupRequest", requests, ["properties"]
+        )
+        self.assertEqual(len(dict_grouped_request.properties), 3)
+        self.assertEqual(
+            list(dict_grouped_request.properties.values())[0].value, "prop1"
+        )
+
```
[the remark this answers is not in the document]
slemasne: Sure, done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                code=code,
+                base_currency="GBP",
+                description="Paper transaction portfolio",
+                created="2020-02-25T00:00:00Z",
+            )
+            return model
+
+        try:
+
+            for code in ["TEST_COM1", "TEST_COM2", "TEST_COM3", "TEST_COM4"]:
```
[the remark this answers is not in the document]
slemasne: sure, done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                cls.api_factory.build(
+                    lusid.api.TransactionPortfoliosApi
+                ).create_portfolio(
+                    scope="test_scope_20200225",
```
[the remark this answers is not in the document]
slemasne: sure, done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
```diff
@@ -0,0 +1,324 @@
...
+                "code": "PortGroupCode",
+                "display_name": "PortGroupDisplayName",
+            },
+            mapping_optional={"values.scope": "Scope", "values.code": "FundCode",},
+            file_type="portfolio_group",
+            property_columns=[],
+            properties_scope=None,
+        )
+
+        if responses.get("portfolio_groups").get("errors") != []:
```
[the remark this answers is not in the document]
slemasne: For the first line, I needed to use .get to check if the key exists. The interpreter throws a KeyError if you try and index on a key which does not exists. 
For the second line, we know the key exists (since the code passed the "if" statement) and therefore we can index as normal using [].

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Sure, done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Updated to len("response error") > 0.

Re the .get approach - I did this to avoid KeyErrors.
Perhaps a better approach would be:
try foo:
except KeyError:
log (bar)

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: As above. 
To avoid KeyError.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Updated to len("response error") > 0.

Re the .get approach - I did this to avoid KeyErrors.
Perhaps a better approach would be:
try foo:
except KeyError:
log (bar)

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: As above.

slemasne: sure, done.

lusidtools/cocoon/utilities.py
slemasne: Thanks!

lusidtools/cocoon/utilities.py
slemasne: sure, done.

lusidtools/cocoon/utilities.py
slemasne: Got it, thanks.

lusidtools/cocoon/config/domain_settings.json
slemasne: Sure. updated.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: sure. updated.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: As above.

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
slemasne: (2) Added all three recommended tests

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Sure, done. 
I checked that the length of the list which contains "successes" is 0.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: You're right. That has been updated.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: Sure, done.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: sure, updated.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: For duplicates:
CreatePortGroup - no errors. The latter request will be an "upsert" on the former request. 
AddPortfolioToGroup - this will throw errors if you provide duplicate portfolios and therefore we take the "set"

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: (1) On second look, it's a weak test - I will remove. 
(2) There should be a test called `test_portfolio_not_exist` for that.

lusidtools/cocoon/config/domain_settings.json
slemasne: `effective_at` is required for add_portfolio_to_group and not create_portfolio_group, so I have not added `effective_at `to the domain config.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
slemasne: added these tests.

lusidtools/cocoon/cocoon.py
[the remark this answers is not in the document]
MikeMcGarry: With this change unfortunately it will have the opposite effect. When you return the error on the failed adding of a portfolio that is the end of the function and no more portfolios will be added. Your test test_08_pg_add_bad_portfolio asserts that with a single bad portfolio an error will be returned. Why don't you try a test with one bad and one good portfolio and see if the good portfolio goes in and the bad one does not?

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: Great nice work! For Providing an empty list you should check that the error message matches your expectation as well. In this case "The provided list of attribute_for_grouping is empty".

For:
- Providing a string of a for a model which doesn't exist e.g. "NotALusidModel"
- Providing an index greater than the length of the list

We need to make sure that we are providing an error which helps the user (even if an internal function) understand what has gone wrong. These two tests only assert that the default unhandled error is raised which is unlikely to help a user understand where the problem can be fixed. You should add some better handling for these cases in your utilities.group_request_into_one function so that you control the error that is raised.

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: Nice work looks great!

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: Nice work looks great!

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: Nice work looks great!

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: Awesome, good job!

tests/integration/cocoon/test_cocoon_portfolio_groups.py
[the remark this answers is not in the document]
MikeMcGarry: Nice one

tests/integration/cocoon/test_cocoon_portfolio_groups.py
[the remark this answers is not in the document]
MikeMcGarry: Sweet, nice work using create_scope_id

tests/integration/cocoon/test_cocoon_portfolio_groups.py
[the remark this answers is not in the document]
MikeMcGarry: .get will still throw a key error unless you specify a default, you would need e.g. responses.get("portfolio_groups", {}).get("errors", []) to get what you desire here.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Nice one

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Nice work, in log_error_requests_title please use an if statement rather than try except.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: As per above

.get will still throw a key error unless you specify a default, you would need e.g. responses.get("portfolio_groups", {}).get("errors", []). You can see this in that StackOverflow post. 

If the try except block you are referring to is the logging please us an if statement instead as per comment above.

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Test Cases:

Can you please update the description of each test? Currently they are the same even for different tests.

Why in most of your tests don't you check that the created/updated portfolio groups contain the portfolios that you expect?

I think you'll find that when you update a portfolio group you will only ever add the first missing portfolio and all others will be ignored.




To do:
- _There is still no test case_ where you have portfolios in different scopes being added to a group or created alongside a group
- You have added a test for trying to update a group when one of the portfolios you are trying to add does not exist _( but need to try with one bad and one or more good to ensure that the good ones go in)_
- You have added a test case (test_10_pg_add_no_new_portfolio) for trying to update a group when there are no new portfolios to add (i.e. all portfolios already added) however this actually does add a portfolio as there are no portfolios in the group you created earlier in the test. 

Done:
- There is no test case for creating a portfolio group with properties **(Nice one)**
- There is no test case for trying to update a group with a duplicate portfolio **(Nice one)**

tests/integration/cocoon/data/port_group_tests/test_1_pg_create_with_portfolios.csv
```diff
@@ -0,0 +1,5 @@
+,PortGroupCode,PortGroupDisplayName,FundCode,
+0,PG_TEST1,Test Portfolio Group,TEST_COM1,
+1,PG_TEST1,Test Portfolio Group,TEST_COM2,
+2,PG_TEST2,Test Portfolio Group,TEST_COM3,
+4,PG_TEST3,Test Portfolio Group,TEST_COM4,
```
thomasdotred: Some of these test files have a leading index column, and others are mssing. Is this necessary/deliberate?

tests/integration/cocoon/data/port_group_tests/test_10_pg_add_no_new_portfolio.csv
```diff
@@ -0,0 +1,2 @@
+PortGroupCode,PortGroupDisplayName,FundCode,
+PG_TEST10,Test Portfolio Group,TEST_COM1,
```
thomasdotred: Some of these test files have a leading index column, and others are mssing. Is this necessary/deliberate?

tests/integration/cocoon/test_utilities.py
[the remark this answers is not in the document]
MikeMcGarry: You will need to check that the error message is as you expect. Your check for the index being greater than the length of the list actually comes after you try and use that index so the IndexError that you get is not the one you wanted to send with a helpful error.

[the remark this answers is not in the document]
MikeMcGarry: Good example here - https://github.com/finbourne/lusid-sdk-python-preview/blob/master/sdk/tests/utilities/test_api_client_factory.py#L48-L50

tests/integration/cocoon/test_cocoon_portfolio_groups.py
MikeMcGarry: Great work! Looks brilliant!

PR #83, opened 2020-03-30
Feature/dev 6075

MikeMcGarry:
# Pull Request Checklist

- [ x] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/docs/CONTRIBUTING.md)
- [ x] Tests pass

# Description of the PR


lusidtools/extract/group_holdings.py
```diff
@@ -0,0 +1,364 @@
...
+                amount=reduce((lambda x,y: x+y), list(map(lambda x: x.cost_portfolio_ccy.amount, value)))
+            ),
+            # Takes the properties from the first value, only allows instrument properties
+            properties={
+                property_key: value for property_key, value in value[0].properties.items() if value.key.split("/")[0] == "Instrument"
+            }
+        )
+    )
+
+    # Use the other details of the first VersionedResourceListOfPortfolioHolding in the response e.g. links etc.
```
rizwansaeed: `links` won't be valid, ditto `requestId`, we should set them to `None` or return a list containing the values for each holding

lusidtools/extract/group_holdings.py
```diff
@@ -0,0 +1,364 @@
...
+    joined_holdings = list(holdings_to_join.values())[0]
```
rizwansaeed: this will throw an exception if an empty dict is passed in

tests/integration/extract/test_group_holdings.py
```diff
@@ -0,0 +1,304 @@
...
+                ]
+            )
+        )
+
+        assert(isinstance(response, lusid.models.PortfolioGroup))
+
+    def check_holdings_correct(self, group_holdings, expected_values):
+
+        # Check that there are the right number of results and they have the correct keys
+        self.assertEqual(len(group_holdings.keys()), len(expected_values.keys()))
```
rizwansaeed: do you need this line given the one below?

tests/integration/extract/test_group_holdings.py
```diff
@@ -0,0 +1,304 @@
...
+        # Iterate over each result
+        for portfolio, holdings in group_holdings.items():
+
+            # Key the result by the instrument name
+            holdings_by_instrument_name = {
+                holding.properties["Instrument/default/Name"].value.label_value: holding for holding in holdings.values
+            }
+
+            # Check that the units and cost are correct
+            [self.assertEqual(int(holding.cost.amount), int(expected_values[portfolio][instrument_name]['cost'])) for instrument_name,
```
rizwansaeed: should be `float` not `int`

rizwansaeed: is it better to use `map()` rather than list comprehension here?

tests/integration/extract/test_group_holdings.py
```diff
@@ -0,0 +1,304 @@
...
+                holding in holdings_by_instrument_name.items()]
+
+            [self.assertEqual(int(holding.units), int(expected_values[portfolio][instrument_name]['units'])) for instrument_name,
```
rizwansaeed: as above

tests/integration/extract/test_group_holdings.py
```diff
@@ -0,0 +1,304 @@
...
+            group_code="ABC12",
+            property_keys=["Instrument/default/Name"],
+            group_by_portfolio=True,
+            num_threads=15
+        )
+
+        expected_results = {
+            f"{self.scope} : Portfolio-X": {
+                "Amazon": {
+                    "cost": "52916721",
```
rizwansaeed: isn't this a `CurrencyAndAmount`?

tests/integration/extract/test_group_holdings.py
```diff
@@ -0,0 +1,304 @@
...
+                    "units": "26598"
+                },
+                "Sainsbury": {
+                    "cost": "188941977",
```
rizwansaeed: this test looks like its cross currency but ignoring currency?

tests/unit/extract/test_group_holdings.py
```diff
@@ -0,0 +1,324 @@
...
+                    effective_from=now
+                ),
+                links=[Link(
+                    relation="MOCK",
+                    href="MOCK",
+                    method="MOCK"
+                )]
+            )
+
+    @parameterized.expand([
```
rizwansaeed: its quite hard to read these tests give the size of the inputs, given the single function call inside the test, I wonder if its better to write them as separate explicitly tests?

tests/unit/extract/test_group_holdings.py
```diff
@@ -0,0 +1,324 @@
...
+                                amount=10000,
+                                currency="AUD"
+                            ),
+                            cost_portfolio_ccy=CurrencyAndAmount(
+                                amount=0,
+                                currency="AUD"
+                            )
+                        )
+                    ]
+                )
```
rizwansaeed: I think you should include another (different) instrument here to show that (A,A,B) becomes (2xA, B)

tests/unit/extract/test_group_holdings.py
```diff
@@ -0,0 +1,324 @@
...
+            }
+        ]
+    ])
```
rizwansaeed: I think we need a test for holdings in different currencies

lusidtools/extract/group_holdings.py
```diff
@@ -0,0 +1,364 @@
...
+    # Update the holdings with the joined holdings
+    joined_holdings.values = holdings
+
+    return {dict_key: joined_holdings}
+
+
+@run_in_executor
+def _get_portfolio_group(api_factory: lusid.utilities.ApiClientFactory, scope: str, code: str,
```
rizwansaeed: is there any reason not to make them all `kwargs`?

lusidtools/extract/group_holdings.py
```diff
@@ -0,0 +1,364 @@
...
+
+    # Extract the portfolio code and scope from the response
+    portfolio_code = response.href.split('/')[7]
+    portfolio_scope = response.href.split('/')[6]
+
+    # Key the response with the unique scope/code combination
+    return {f"{portfolio_scope} : {portfolio_code}": response}
+
+
+async def _get_holdings_for_group_recursive(
```
rizwansaeed: make `kwargs`?

lusidtools/extract/group_holdings.py
```diff
@@ -0,0 +1,364 @@
...
+
+        # Join together the holdings across all the sub-groups
+        joined_sub_group_holdings = _join_holdings(sub_group_holdings_keyed, group_by_portfolio, dict_key="SubGroupHoldings")
+
+        # Join and return the joined sub-group holdings with the joined portfolio holdings
+        return _join_holdings({**joined_sub_group_holdings, **joined_portfolio_holdings}, group_by_portfolio,
+                              dict_key=f"{group_scope} : {group_code}")
+
+
+def get_holdings_for_group(
```
rizwansaeed: make `kwargs`?

tests/integration/extract/test_group_holdings.py
MikeMcGarry: Yes, testing currency now as well.

lusidtools/extract/group_holdings.py
MikeMcGarry: I have removed VersionedResourceListOfPortfolioHolding altogether in favour of just a list of PortfolioHolding

lusidtools/extract/group_holdings.py
MikeMcGarry: Removed due to using a list of PortfolioHolding instead

lusidtools/extract/group_holdings.py
MikeMcGarry: On second thought you are right this is much nicer. Moved to using kwargs. There still needs to be a filtering step before calling LUSID but it is much cleaner this way.

tests/unit/extract/test_group_holdings.py
MikeMcGarry: Done.

tests/unit/extract/test_group_holdings.py
MikeMcGarry: Found a much nicer solution, thanks!

PR #91, opened 2020-04-15
DEV-6172: Refactor create_scope_id into shell function

thomasdotred:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

This PR encapsulates the time-based ID creation into another function, not specific to scope. Create_scope_id is left as a shell for compatibility with existing usages

### DIFF
--- a/lusidtools/cocoon/utilities.py
+++ b/lusidtools/cocoon/utilities.py
@@ -1317,6 +1317,46 @@ def strip_whitespace(df: pd.DataFrame, columns: list) -> pd.DataFrame:
     return stripped_df
 
 
+def generate_time_based_unique_id(time_generator: None):
+    """
+    Generates a unique ID based on the time since epoch.
+
+    Parameters
+    ----------
+    time_generator
+        Any class that has a .time() method on it which produces time since 1970 in seconds
+
+    Returns
+    -------
+    uid : str
+        Unique, time based ID
+
+    """
+
+    if time_generator is None or isinstance(time_generator, types.ModuleType):
+        time_generator = default_time
+
+    elif getattr(time_generator, "time", None) is None or not isinstance(
+        getattr(time_generator, "time"), types.MethodType
+    ):
+        raise AttributeError(
+            "The provided time_generator does not have a method called time"
+        )
+
+    # Get the current time since epoch
+    current_time = time_generator.time()
+
+    if not isinstance(current_time, int) and not isinstance(current_time, float):
+        raise ValueError(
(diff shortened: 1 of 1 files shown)


lusidtools/cocoon/utilities.py
```diff
@@ -19,6 +19,7 @@
 import time as default_time
 from lusidtools.cocoon.validator import Validator
 import types
+import uuid
```
slemasne, line 22: Where is this used in the code? It looks like you're generating your own uid from the timestamp hex?

lusidtools/cocoon/utilities.py
```diff
@@ -1354,8 +1355,26 @@ def create_scope_id(time_generator=None):
     # Multiply by 7 to get value to 100s of nano seconds
     timestamp = hex(int(test * 10000000.0))
```
slemasne, line 1356: Perhaps change the `test` variable name to `current_time`

lusidtools/cocoon/utilities.py
```diff
@@ -1354,8 +1355,26 @@ def create_scope_id(time_generator=None):
...
     # Create the scope id by joining the hex representation with dashes every 4 characters
-    scope_id = "-".join(timestamp[i : i + 4] for i in range(2, len(timestamp), 4))
-    return scope_id
+    uid = "-".join(timestamp[i : i + 4] for i in range(2, len(timestamp), 4))
```
slemasne, line 1358: The `load_from_data_frame` function doesnt like `-` and it converts them to `-` but thats an issue with `load_from_data_frame` - we should probably fix that function to allow `-` as LUSID permits `-`

thomasdotred: Um whats the difference between the two `-` above? I cant see in the formatting!

lusidtools/cocoon/utilities.py
thomasdotred: Changed

lusidtools/cocoon/utilities.py
thomasdotred: removed

lusidtools/cocoon/utilities.py
slemasne: Underscore (_) versus hyphen (-).
The function like underscore.

thomasdotred: Ah ok, I must have miss-read as a hyphen

PR #92, opened 2020-04-15
DEV-7563: Add check for empty mapping string

thomasdotred:
DEV-7563: Add Check for empty string

# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

This PR Fixes an underhanded error when an empty string is passed in the mapping

### DIFF
--- a/lusidtools/cocoon/utilities.py
+++ b/lusidtools/cocoon/utilities.py
@@ -836,6 +836,11 @@ def handle_nested_default_and_column_mapping(
                 )
 
         elif isinstance(value, str):
+
+            if len(value) == 0:
+                raise IndexError(f"Unspecified mapping field: {key}. Please assign a value or remove this from the "
+                                 f"mapping")
+
             if value[0] != constant_prefix:
                 mapping_updated[key] = value
             else:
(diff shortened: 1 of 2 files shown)


lusidtools/cocoon/utilities.py
```diff
@@ -836,6 +836,11 @@ def handle_nested_default_and_column_mapping(
                 )
 
         elif isinstance(value, str):
+
+            if len(value) == 0:
+                raise IndexError(f"Unspecified mapping field: {value}. Please assign a value or remove this from the "
```
MikeMcGarry, line 841: I think you want this to be {key} instead of value as this error will only ever show nothing.

thomasdotred: Agreed

tests/unit/cocoon/test_utilities.py
```diff
@@ -1095,8 +1099,25 @@ def test_handle_nested_default_and_column_mapping_success(
 
     @parameterized.expand(
         [
+            # Test empty string
+            [
+                "Test empty string",
```
rizwansaeed, line 1104: I think it would be better to pull the DataFrame construction out into a separate parameterised function to make it easier to read the test cases, not a blocker for this PR but maybe raise a Mira

PR #95, opened 2020-04-20
DEV-7674: Fix default_fx_model() remapping bug

thomasdotred:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR
default_FX_model() takes FX transactions represented as two transactions (buy and sell legs) and merges them into a single line. It then updates the mapping for the LUSID fields with the updated column names in the new merged dataframe. 

Currently the total consideration fields are re-mapped to be buy fields, and vice versa for transaction fields. This PR fixes this so that total consideration is remapped to the sell columns


lusidtools/cocoon/utilities.py
```diff
@@ -1470,14 +1470,14 @@ def remap_after_merge(mapping: dict, buy_suffix: str, sell_suffix: str) -> dict:
...
-    buy_fields = [
+    sell_fields = [
         "total_consideration.amount",
         "total_consideration.currency",
         "settlement_currency",
     ]
 
     # currencies and amounts leaving the portfolio i.e. sell
-    sell_fields = ["units", "transaction_currency"]
+    buy_fields = ["units", "transaction_currency"]
```
slemasne, line 1480: How are we certain these are always the `buy` fields?
For example, consider a `FxBuy` of GBP/USD.
Here the units and **transaction_currency** are **indeed** the buy fields.
But what if we're doing a `SellFx` of GBP/USD.
Then the **units** and **transaction_currency** are the sell fields.

thomasdotred: in  the latter case, wouldn't the units and transaction still need to be mapped the same, but in the configuration they would have opposite sides?

slemasne: So in the standard LUSID canonical transaction types, `SellFx` has the following side configuration:

- CashCommitment on Side1 with a -1 direction
- CashCommitment on Side2 with a 1 direction

In plain English, that means we're:

- Decreasing the cash of the **transaction_currency** by the **units**
- Increasing the cash in the settlement_currency by the **total_consideration.amount**

Therefore labelling **transaction_currency** and **units** as buy_fields is not always correct?

(apologies if I am missing something)

PR #100, opened 2020-05-06
WIP: DEV-7805: first cut of order creation tool

timbarrass:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

New lusidtools tool to create orders. Reads an order per row from a csv file.

### DIFF
--- a/lusidtools/lpt/create_orders.py
+++ b/lusidtools/lpt/create_orders.py
@@ -1,0 +1,94 @@
+import pandas as pd
+
+from lusidtools.lpt import lpt
+from lusidtools.lpt import lse
+from lusidtools.lpt import stdargs
+
+NAME = "name"
+TOOLNAME = "orders_create"
+TOOLTIP = "Create Orders"
+
+
+def parse(extend=None, args=None):
+    return (
+        stdargs.Parser("Upsert Orders", ["filename", "limit", "test"])
+        .add("input", nargs="+")
+        .add(
+            "--mappings",
+            nargs="+",
+            help="column name mappings of the form TICKER=col1 etc",
+        )
+        .add(
+            "--identifiers",
+            nargs="+",
+            default=["ClientInternal", "Figi"],
+            help="Identifier types provided",
+        )
+        .extend(extend)
+        .parse(args)
+    )
+
+
+def process_args(api, args):
+    aliases = {
+        "CINT": "ClientInternal",
+        "FIGI": "Figi",
+        "RIC": "P:Instrument/default/RIC",
+        "TICKER": "P:Instrument/default/Ticker",
+        "ISIN": "P:Instrument/default/Isin",
+    }
+
+    if args.input:
+        df = pd.concat(
+            [lpt.read_input(input_file, dtype=str) for input_file in args.input],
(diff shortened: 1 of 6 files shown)


lusidtools/lpt/create_orders.py
```diff
@@ -0,0 +1,94 @@
...
+                ),
+                inplace=True,
+            )
+
+        prop_keys = [col for col in df.columns.values if col.startswith("P:")]
+
+        identifiers = [col for col in df.columns.values if col in args.identifiers]
+
+        # Identifiers have to be unique
+        df = df.drop_duplicates(identifiers)
```
MikeMcGarry, line 64: These lines are an exact copy of lines 18-67 in create_instr.py. Is it possible to move both of these functions into a common module that can be imported by both create_instr.py and create_orders.py?

lusidtools/lpt/create_orders.py
```diff
@@ -0,0 +1,94 @@
...
+from lusidtools.lpt import stdargs
+
+NAME = "name"
+TOOLNAME = "orders_create"
+TOOLTIP = "Create Orders"
+
+
+def parse(extend=None, args=None):
+    return (
+        stdargs.Parser("Upsert Orders", ["filename", "limit", "test"])
```
MikeMcGarry, line 14: Should this match the Tooltip? i.e. "Create Orders" in both or "Upsert Orders" in both.

lusidtools/lpt/create_orders.py
```diff
@@ -0,0 +1,94 @@
...
+        order_request=api.models.OrderSetRequest(
+            order_requests=[
+                api.models.OrderRequest(
+                    id=api.models.ResourceId(row["id.scope"], row["id.code"]),
+                    side=row["side"],
+                    quantity=row["quantity"],
+                    order_book_id=api.models.ResourceId(row["orderBookId.scope"], row["orderBookId.code"]),
+                    portfolio_id=api.models.ResourceId(row["portfolioId.scope"], row["portfolioId.code"]),
+                    instrument_identifiers={'Instrument/default/' + identifier: row[identifier] for identifier in identifiers},
+                    properties={key[2:]: api.models.ModelProperty(key[2:], api.models.PropertyValue(row[key])) for key in prop_keys if pd.notna(row[key])}
```
MikeMcGarry, line 79: We have two different implementations for loading data from a file. One in the lpt folder which is designed to be called from the command line and one in the cocoon folder which is designed to be called using code. For this addition, what is the use case? Is there scope to create it using the cocoon implementation (which has more flexibility on names in the file and takes a configuration to map from the file to LUSID) or is there a pressing need to get this live with this hard coded implementation?

MikeMcGarry: As discussed will go with the lpt implementation for now.

lusidtools/lpt/create_orders.py
timbarrass: Have factored out most into a new method in lpt.py

lusidtools/lpt/create_orders.py
timbarrass: Done

PR #103, opened 2020-05-15
DEV-7989 Fixed test after change from DEV-7620 Instead of the scope a…

MikeMcGarry:
…nd code of a source portfolio being returned as a / seperated string in Holding/default/SourcePortfolioId, they are now populated in separate properties.  Scope: Holding/default/SourcePortfolioScope Code:  Holding/default/SourcePortfolioId  Need to update the test which checks that the LUSID Tools function get_holdings_for_group gets the same result as you would from LUSID via the Get Holdings for Group API to reflect this change.

# Pull Request Checklist

- [x ] Tests pass

# Description of the PR


MikeMcGarry:
Updated to use assertSetEquals

PR #107, opened 2020-05-27
DEV-8199: Remove column in identify_cash_items where remove == True

thomasdotred:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

### DIFF
--- a/tests/unit/cocoon/test_utilities.py
+++ b/tests/unit/cocoon/test_utilities.py
@@ -1340,7 +1340,7 @@ class CocoonUtilitiesTests(unittest.TestCase):
             ),
         ]
     )
-    def test_scale_quote_of_type(self, _, data, ground_truth, scale_factor):
+    def test_scale_quote_of_type(self, _, data, expected_value, scale_factor):
         df = pd.DataFrame(data, columns=["name", "type", "price"])
         mapping = {
             "quotes": {
@@ -1365,7 +1365,7 @@ class CocoonUtilitiesTests(unittest.TestCase):
         result, mapping = cocoon.utilities.scale_quote_of_type(df=df, mapping=mapping)
 
         [
-            self.assertEqual(ground_truth[index], row["__adjusted_quote"])
+            self.assertEqual(expected_value[index], row["__adjusted_quote"])
             for index, row in result.iterrows()
         ]
 
@@ -1441,7 +1441,6 @@ class CocoonUtilitiesTests(unittest.TestCase):
                     },
                     "implicit": "internal_currency",
                 },
-                False,
                 ["GBP_EXP", "GBP_EXP", "USD_EXP", "USD_EXP"],
             ),
             (
@@ -1456,7 +1455,6 @@ class CocoonUtilitiesTests(unittest.TestCase):
                         },
                     }
                 },
-                False,
                 ["GBP_IMP", "GBP_IMP", "USD_IMP", "USD_IMP"],
             ),
             (
@@ -1472,7 +1470,6 @@ class CocoonUtilitiesTests(unittest.TestCase):
                     },
                     "implicit": "internal_currency",
                 },
-                False,
                 ["GBP_IMP", "GBP_IMP", "USD_IMP", "USD_EXP"],
             ),
             (
@@ -1483,7 +1480,6 @@ class CocoonUtilitiesTests(unittest.TestCase):
                     },
                     "implicit": "internal_currency",
                 },
-                False,
                 [],
             ),
             (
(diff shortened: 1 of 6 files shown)


tests/unit/cocoon/test_utilities.py
```diff
@@ -1674,10 +1674,14 @@ def test_identify_cash_items(self, _, cash_flag, remove_cash_items, ground_truth
         dataframe, mappings_test = identify_cash_items(
             dataframe, mappings, file_type, remove_cash_items
         )
-
-        self.assertEqual(
-            ground_truth, list(dataframe["__currency_identifier_for_LUSID"])
-        )
+        if not remove_cash_items:
```
rizwansaeed, line 1677: you should avoid having if logic in tests, these should be split into 2 tests

tests/unit/cocoon/test_utilities.py
```diff
@@ -1587,24 +1587,44 @@ def test_identify_cash_items_failed(
...
-                    "implicit": "internal_currency",
-                },
-                False,
-                ["GBP_EXP", "GBP_EXP", "USD_EXP", "USD_IMP"],
-            ),
+                ["GBP_EXP", "GBP_EXP", "USD_EXP", "USD_EXP", None],
+            )
+        ]
+    )
+    def test_identify_cash_items_remove_false(self, _, cash_flag, remove_cash_items, ground_truth):
```
rizwansaeed, line 1594: couple of things:
1) its not clear from the name what this test is meant to be doing, i think it should be renamed to something like `test_remove_cash_items_with_identify_cash_items`
2) i think `ground_truth` is your expected values, rename to something like `expected_values`
3) `remove_cash_items` doesn't look like it needs to be passed in and can be set in the call - its `False` for all

tests/unit/cocoon/test_utilities.py
```diff
@@ -1587,24 +1587,44 @@ def test_identify_cash_items_failed(
...
+            "Currency"
+        ] = "__currency_identifier_for_LUSID"
+
+        dataframe = pd.DataFrame(data)
+
+        dataframe, mappings_test = identify_cash_items(
+            dataframe, mappings, file_type, remove_cash_items
+        )
+
+        self.assertEqual(
```
rizwansaeed, line 1620: might be better to use `assertListEqual`

tests/unit/cocoon/test_utilities.py
```diff
@@ -1645,41 +1665,38 @@ def test_identify_cash_items_failed(
                     "implicit": "internal_currency",
                 },
                 True,
-                [],
+                ["inst5"],
             ),
         ]
     )
-    def test_identify_cash_items(self, _, cash_flag, remove_cash_items, ground_truth):
+    def test_identify_cash_items_remove_true(self, _, cash_flag, remove_cash_items, ground_truth):
```
rizwansaeed, line 1672: as above for naming and removing `remove_cash_items` parameter

tests/unit/cocoon/test_utilities.py
thomasdotred: 3. ✔️

thomasdotred: 2. ✔️

thomasdotred: 1, ✔️ :
`test_identify_cash_items_with_remove_cash_items()`
`test_identify_cash_items_without_remove()`

tests/unit/cocoon/test_utilities.py
thomasdotred: `test_identify_cash_items_without_remove()` as above

tests/unit/cocoon/test_utilities.py
thomasdotred: Made change, also used assertDictEquals

tests/unit/cocoon/test_utilities.py
rizwansaeed: there are still some "ground truths" in there, can you rename to "expected values" or something equivalent

rizwansaeed:
i think the sonar build has picked up a potential bug, can you fix

thomasdotred:
> i think the sonar build has picked up a potential bug, can you fix

The code smells are where I used a list comprehension, but don't use the iterating variable. I have replaces `i` with `_` and sonar cloud has stopped complaining about it.

thomasdotred:
The code duplications are two assertions that are the  same across two of the tests. I think that abstracting these tests will reduce readability of the tests so I have left it out

PR #108, opened 2020-06-01
adding black to travis

slemasne:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


.travis.yml
```diff
@@ -3,30 +3,46 @@ services:
 - docker
 install:
 - pip install twine wheel pyOpenSSL
-script:
-- |
+- pip install -r requirements.txt
 
-  export last_commit_author=`git log -1 --pretty=format:%ae`
+branches:
```
rizwansaeed, line 8: do we need this? we want it to run on all branches but only commit where it is not master - which I think you have covered in the check later

.travis.yml
```diff
@@ -3,30 +3,46 @@ services:
...
-      -e FBN_TOKEN_URL="$fbn_token_url" \
-      -e FBN_USERNAME="$fbn_username" \
-      -e FBN_PASSWORD="$fbn_password" \
-      -e FBN_CLIENT_ID="$fbn_clientid" \
-      -e FBN_CLIENT_SECRET="$fbn_client_secret" \
-      -e FBN_LUSID_API_URL="$fbn_lusid_api_url" \
-      -e FBN_APP_NAME="$fbn_app_name" \
-      -v $(pwd):/usr/src/ \
-      finbourne/lusid-python-tools-test
+        export last_commit_author=`git log -1 --pretty=format:%ae`
```
rizwansaeed, line 23: I don't think you need this line any more

ci_tools/new_commit_with_black.sh
```diff
@@ -0,0 +1,10 @@
+black .
```
rizwansaeed, line 1: I wonder if its better to just have black run in black_formatter_check.sh and update if needed and then in this script commit back any changes that it made. Reason being if we have it run in 2 places there is scope for them to diverge wrt to parameters. So it looks something like

1) run black and update source in repo
2) if branch != master commit repo to branch
3) run tests
4) publish

ci_tools/new_commit_with_black.sh
```diff
@@ -0,0 +1,10 @@
...
+if [ -n "$(git status --porcelain)" ]; then 
+git config --local user.email "action@github.com" 
+git config --local user.name "GitHub Action"
```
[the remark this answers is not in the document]
slemasne: Thanks. Updated.

ci_tools/new_commit_with_black.sh
```diff
@@ -0,0 +1,10 @@
+black .
+if [ -n "$(git status --porcelain)" ]; then 
+git config --local user.email "action@github.com"
```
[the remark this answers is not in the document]
slemasne: Thanks. What is the email? (I dont have access to the account to check).
finbourne-bot-public@github.com?

.travis.yml
slemasne: Agreed. Removed.

ci_tools/new_commit_with_black.sh
slemasne: Thanks.
I've changed the approach slightly.
There was an issue with the build on the feature branch.

**Old approach**

On feature:
(1) Run black and commit directly to feature
(2) Do package tests
(3) Check that code is black compliance

The issue with this approach - step #3 would always fails as Travis pulls the commit before black ran. I think it might cache a commit from when you start the build. I searched online but could not find a solution.

**New approach (which works)**

On feature:
(1) Run black and commit directly to feature
(2) Run package tests

On master PR
(1) Run package test
(2) Check that PR is black compliant

This approach works as the master build pulls the latest commit made with black formatting.

.travis.yml
slemasne: Removed.

ci_tools/new_commit_with_black.sh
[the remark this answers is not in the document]
rizwansaeed: use engineering@finbourne.com

ci_tools/new_commit_with_black.sh
rizwansaeed: on  master, it might be worth running the black check first as it will be quicker and you want to fail fast

ci_tools/new_commit_with_black.sh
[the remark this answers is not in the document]
slemasne: Thanks. Done.

ci_tools/new_commit_with_black.sh
slemasne: Thanks. Done.

PR #111, opened 2020-06-04
DEV-6172: Make Time Generator more unique

thomasdotred:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


lusidtools/cocoon/utilities.py
```diff
@@ -1361,7 +1363,7 @@ def generate_time_based_unique_id(time_generator: None):
     timestamp = hex(int(current_time * 10000000.0))
     # Create the scope id by joining the hex representation with dashes every 4 characters
     uid = "-".join(timestamp[i : i + 4] for i in range(2, len(timestamp), 4))
-    return uid
+    return uid + str(uuid.uuid4())
```
MikeMcGarry, line 1366: Not sure on whether this actually makes the id any more unique, let's verify there is an issue with the existing implementation then go from there.

lusidtools/cocoon/utilities.py
```diff
@@ -1361,7 +1363,7 @@ def generate_time_based_unique_id(time_generator: None):
     timestamp = hex(int(current_time * 10000000.0))
```
MikeMcGarry, line 1363: Can increase this to * 1000000000 (microsends) instead of 10000000 (rounded to 100s of microseconds) if they are running too close to each other.

lusidtools/cocoon/utilities.py
MikeMcGarry: Verified, that this is an issue. Can you please move creating the uuid and appending to the create_scope_id function?

lusidtools/cocoon/utilities.py
```diff
@@ -1380,7 +1382,7 @@ def create_scope_id(time_generator=None):
         Scope identifier
     """
 
-    return generate_time_based_unique_id(time_generator)
+    return generate_time_based_unique_id(time_generator) + str(uuid.uuid4())
```
MikeMcGarry, line 1385: Can this be used as-is in most operations? Or does it always need to be truncated to be used as an id in LUSID because it is too long? I believe the limit is 64 characters for scopes and codes.

PR #119, opened 2020-06-08
amending bump version to prevent duplicate version commits

slemasne:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

### DIFF
--- a/ci_tools/bumpversion.sh
+++ b/ci_tools/bumpversion.sh
@@ -1,5 +1,27 @@
-bump2version patch --message "Bump version: {current_version} to {new_version} [ci skip]"
-git config --local user.email "engineering@finbourne.com"
-git config --local user.name "finbourne-bot-public"
-git push https://finbourne-bot-public:${gh_token}@github.com/finbourne/lusid-python-tools.git HEAD:$TRAVIS_BRANCH
-echo "Making commit with incremented version"
+#!/bin/bash
+
+last_commit_message=$(git log -1 HEAD --pretty=format:%s)
+
+# Bump the version up
+
+if [[ "$last_commit_message" == "Bump version"* ]]; then
+        echo "Skipping version increment (no commits found)"
+else
+
+	# Find next version
+
+	api_version=$(cat __version__.py | grep __version__ |  awk '{split($0, a, "="); print a[2]}' | tr -d ' "')
+	IFS='.' read -a version_split <<< "${api_version}"
+	current_patch=${version_split[2]}
+	new_patch=$(echo "$current_patch + 1" | tr -d $'\r' | bc)
+	new_version="${version_split[0]}.${version_split[1]}.$new_patch"
+    echo "Bumping packaging patch version from $api_version to $new_version"
+
+    # Bump the version in project and make commit back to master
+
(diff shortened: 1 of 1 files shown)


ci_tools/bumpversion.sh
```diff
@@ -1,5 +1,16 @@
...
-git config --local user.email "engineering@finbourne.com"
-git config --local user.name "finbourne-bot-public"
-git push https://finbourne-bot-public:${gh_token}@github.com/finbourne/lusid-python-tools.git HEAD:$TRAVIS_BRANCH
-echo "Making commit with incremented version"
\ No newline at end of file
+#!/bin/bash
+
+last_commit_message=$(git log -1 HEAD --pretty=format:%s)
+
+echo "Last commit message: $last_commit_message"
```
rizwansaeed, line 5: this looks like debug, I don't think we need to print it out

ci_tools/bumpversion.sh
```diff
@@ -1,5 +1,16 @@
...
+
+if [[ "$last_commit_message" == "Bump version"* ]]; then
+        echo "last commit is a bump version commit... skipping..."
```
rizwansaeed, line 8: The pedant in me dislikes the inconsistent casing in the log messages, sorry!

I think change to something like "Skipping version increment (no commits found)"

ci_tools/bumpversion.sh
```diff
@@ -1,5 +1,16 @@
...
+else
+        echo "Bumping packaging patch version by one"
```
rizwansaeed, line 10: Maybe just a simple "Incrementing patch version to $new_version"

ci_tools/bumpversion.sh
slemasne: Sure, removed.

ci_tools/bumpversion.sh
slemasne: Sure. done.

ci_tools/bumpversion.sh
slemasne: I'm not sure the Travis environment will know the $new_version variable? We could always scrape this from PyPi?

rizwansaeed: if we don't have it at hadn't then I think its ok to leave

slemasne: See latest commit. We can take the version from `__version__.py` and then increment in the shell script.

ci_tools/bumpversion.sh
```diff
@@ -1,5 +1,24 @@
...
-git push https://finbourne-bot-public:${gh_token}@github.com/finbourne/lusid-python-tools.git HEAD:$TRAVIS_BRANCH
-echo "Making commit with incremented version"
\ No newline at end of file
+#!/bin/bash
+
+last_commit_message=$(git log -1 HEAD --pretty=format:%s)
+
+# Find next version
+
+api_version=$(cat __version__.py | grep __version__ |  awk '{split($0, a, "="); print a[2]}' | tr -d ' "')
```
rizwansaeed, line 7: i think this block can be moved to line 18 (where it is used)

slemasne: Done

PR #122, opened 2020-06-17
DEV-6172: Add UUID generator to Create scope ID

thomasdotred:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

### DIFF
--- a/tests/unit/cocoon/test_utilities.py
+++ b/tests/unit/cocoon/test_utilities.py
@@ -1702,20 +1702,14 @@ class CocoonUtilitiesTests(unittest.TestCase):
 
         self.assertTrue(df_true.equals(df_test))
 
-    @parameterized.expand(
-        [
-            [
-                "2019-11-27T11:08:38Z",
-                MockTimeGenerator(current_datetime=1574852918),
-                "37f3-342f-823f-00",
-            ]
-        ]
-    )
-    def test_create_scope_id_success(self, _, time_generator, expected_outcome):
 
+    def test_create_scope_id_success(self):
+        time_generator = MockTimeGenerator(current_datetime=1574852918)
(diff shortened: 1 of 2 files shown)


tests/unit/cocoon/test_utilities.py
```diff
@@ -1711,10 +1713,16 @@ def test_strip_whitespace(self, _, val_1, val_2):
     )
     def test_create_scope_id_success(self, _, time_generator, expected_outcome):
```
rizwansaeed, line 1714: this doesn't need to be `parameterized`

tests/unit/cocoon/test_utilities.py
```diff
@@ -1711,10 +1713,16 @@ def test_strip_whitespace(self, _, val_1, val_2):
...
-        scope_id = create_scope_id(time_generator)
+        scope_id = create_scope_id(time_generator=time_generator)
 
         self.assertEqual(first=expected_outcome, second=scope_id)
 
+    def test_create_scope_id_uuid_success(self):
+
+        scope_id = create_scope_id(use_uuid=True)
+
+        self.assertTrue(uuid.UUID(scope_id))
```
rizwansaeed, line 1724: what is this asserting? looks like its just creating a new UUID? i think  you need to assert on `scope_id` and check it 'looks' like a uuid

thomasdotred: if `scope_id` is not a valid uuid, `uuid.UUID` returns the Value error below

```python
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Python38\lib\uuid.py", line 169, in __init__
    raise ValueError('badly formed hexadecimal UUID string')
ValueError: badly formed hexadecimal UUID string
```

PR #126, opened 2020-06-22
DEV-8525 Remove deprecated numpy.datetime64 keyword arguments format …

MikeMcGarry:
…and utc. numpy.datetime64 is always assumed to be datetime naive. Keep in mind that this is also an expiremental API from numpy

# Pull Request Checklist

- [x ] Tests pass

# Description of the PR


rizwansaeed:
good spot, I wonder if its worth pinning the versions completely?

PR #128, opened 2020-06-23
Add ReferencePortfolio to LPT

slemasne:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

This PR provides users ability to create Reference Portfolios through LPT.

TODO:

Add ability to set constituents through LPT
Add ability to update properties on existing portfolio


lusidtools/cocoon/cocoon.py
```diff
@@ -297,6 +297,61 @@ def load_portfolio_batch(
...
+            The code of the reference portfolio to create
+        kwargs
+
+        Returns
+        -------
+        lusid.models.ReferencePortfolio
+            the response from LUSID
+        """
+
+        if "scope" not in list(kwargs.keys()):
```
rizwansaeed, line 328: i think this can just be `if "scope" not in kwargs`

lusidtools/cocoon/cocoon.py
```diff
@@ -297,6 +297,61 @@ def load_portfolio_batch(
...
+            raise KeyError(
+                "You are trying to load a reference portfolio without a scope, please ensure that a scope is provided."
+            )
+
+        if "code" not in list(kwargs.keys()):
```
rizwansaeed, line 333: as above

lusidtools/cocoon/cocoon.py
```diff
@@ -297,6 +297,61 @@ def load_portfolio_batch(
...
+        except lusid.exceptions.ApiException as e:
+            if e.status == 404:
+                return api_factory.build(
+                    lusid.api.ReferencePortfolioApi
+                ).create_reference_portfolio(
+                    scope=kwargs["scope"],
+                    create_reference_portfolio_request=reference_portfolio_batch[0],
+                )
+            else:
+                return e
```
rizwansaeed, line 353: should this be a `raise`

tests/integration/cocoon/test_reference_portfolios.py
```diff
@@ -0,0 +1,132 @@
...
+                ],
+            )
+
+            property_keys_from_params = [
+                f"Portfolio/{self.scope}/{code}"
+                for code in property_columns
+                if len(property_columns) > 0
+            ]
+
+            self.assertListEqual(
```
rizwansaeed, line 130: i don't think order is guaranteed, this may need to be converted to a `set` and `assertSetEquals`

tests/integration/cocoon/test_reference_portfolios.py
```diff
@@ -0,0 +1,132 @@
...
+                "data/reference-test.csv",
+                {"code": "FundCode", "display_name": "display_name"},
+                {
+                    "description": "description",
+                    "created": "created",
+                    "base_currency": "base_currency",
+                },
+                ["strategy", "custodian"],
+                [],
+            ],
```
rizwansaeed, line 49: need a test case for failure case e.g. not supplying all mandatory params and maybe also adding duplicates in the source DataFrame

lusidtools/cocoon/cocoon.py
slemasne: Sure, done.

lusidtools/cocoon/cocoon.py
slemasne: Yes, updated.

tests/integration/cocoon/test_reference_portfolios.py
slemasne: Sure.
Or we could use assertCountEqual.
I have just read this will compare two lists regardless of order which is what we want. 
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

rizwansaeed: that would work too, I thought assertCountEqual would, well check the counts were the same but if checks the values too then perfect!

slemasne: Thanks. It appears to check **count** and **contents** which is what we want (agreed name is slightly misleading).

tests/integration/cocoon/test_reference_portfolios.py
slemasne: I added:

(1) Test for mandatory params
(2) Test for duplicates - duplicates are handled by a function in the `cocoon.py` so passing duplicates should not return an error.

tests/integration/cocoon/test_reference_portfolios.py
```diff
@@ -0,0 +1,180 @@
...
+        ]
+    )
+    def test_load_from_data_frame_a_portfolios_failure(
+        self,
+        _,
+        file_name,
+        mapping_required,
+        mapping_optional,
+        property_columns,
+        expected_outcome,
```
rizwansaeed, line 72: doesn't look like this is used

slemasne: Good catch. Removed.

lusidtools/cocoon/cocoon.py
MikeMcGarry: No we don't want to raise an exception here. That is handled further up the stack. We just want to return an exception here. If we raise an exception here then if a single Portfolio fails to be created on the load the entire load will fail.

tests/integration/cocoon/test_reference_portfolios.py
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    "description": "description",
+                    "created": "created",
+                    "base_currency": "base_currency",
+                },
+                [],
+            ],
+        ]
+    )
+    def test_load_from_data_frame_a_portfolios_failure(
```
MikeMcGarry, line 62: This should be renamed to indicate it is looking for successes, not failures.

lusidtools/cocoon/cocoon.py
rizwansaeed: for now keep it consistent with the others (return an exception), but I don't think a function should have two possible return types, we need to review this later

tests/integration/cocoon/test_reference_portfolios.py
slemasne: Agreed. Good catch. Updated.

lusidtools/cocoon/cocoon.py
slemasne: Thanks. Had a chat with Riz. We changed back to `return`. We can discuss implementation later.

PR #136, opened 2020-07-07
DEV-8665 Rename deprecated aggregation function

rizwansaeed:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

### DIFF
--- a/tests/integration/lpt/test_lpt.py
+++ b/tests/integration/lpt/test_lpt.py
@@ -233,3 +233,27 @@ class LptTests(unittest.TestCase):
             lambda left: self.fail(lpt.display_error(left)),
             lambda right: self.assertIsInstance(right, DataFrame),
         )
+
+    def test_aggregate_group_holdings(self):
+
+        if not self.api.call.get_portfolio_group(scope="Finbourne-Examples", code="US-FI").is_right():
+            self.skipTest("missing target portfolio group")
+
+        qah.process_args(
+            self.api,
+            qah.parse(
+                args=[
+                    "Finbourne-Examples",
+                    "US-FI",
+                    "2020-01-01",
+                    "--group",
+                    "--pricing-scope",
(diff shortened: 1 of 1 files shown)


tests/integration/lpt/test_lpt.py
```diff
@@ -233,3 +231,27 @@ def test_aggregate_holdings(self):
...
+
+        if not self.target_portfolios_exist():
+            self.skipTest("missing target portfolios")
+
+        qah.process_args(
+            self.api,
+            qah.parse(
+                args=[
+                    "Finbourne-Examples",
+                    "US-FI",
```
MikeMcGarry, line 245: Should there be a statement like self.target_portfolios_exist() to assert that this group exists?

tests/integration/lpt/test_lpt.py
```diff
@@ -233,3 +231,27 @@ def test_aggregate_holdings(self):
...
+                    "--group",
+                    "--pricing-scope",
+                    "Finbourne-Examples",
+                    "--recipe",
+                    "FinbourneExamplesRecipeMidThenBid",
+                ]
+            ),
+        ).match(
+            lambda left: self.fail(lpt.display_error(left)),
+            lambda right: self.assertIsInstance(right, DataFrame),
```
MikeMcGarry, line 256: Am I right in saying this request could fail and the test would still pass?

rizwansaeed: no, `process_arg` returns an `Either`, if the function fails it returns a `left` value and if successful it returns `right`. `match` then matches on these values, so if the call fails it will match on `left` and `self.fail`

tests/integration/lpt/test_lpt.py
rizwansaeed: good spot, added

PR #138, opened 2020-07-21
Feature/se 100.format reference portfolios response

jyaffe:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

SE-100: Added a formatting function in cocoon_printer for reference portfolios. Tests also added in test_cocoon_printer.
https://finbourne.atlassian.net/browse/SE-100

### DIFF
--- a/tests/unit/cocoon/test_cocoon_printer.py
+++ b/tests/unit/cocoon/test_cocoon_printer.py
@@ -179,6 +179,13 @@ responses_no_success_field = {
     "reference_portfolios": {"errors": [api_exception for _ in range(2)],},
 }
 
+def assert_responses(self, num_items, expected_value, succ=None, err=None, failed=None):
+        for response in [succ, err, failed]:
+            if response is not None:
+                self.assertEqual(num_items, len(response))
+                for index, row in response.iterrows():
+                    self.assertEqual(expected_value[f"{response}"][index], row[response.columns[0]])
+
 
 class CocoonPrinterTests(unittest.TestCase):
     @classmethod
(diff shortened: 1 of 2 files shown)


tests/unit/cocoon/test_cocoon_printer.py
```diff
@@ -363,6 +383,33 @@ def test_format_holdings_response_success(
...
+        ]
+    )
+    def test_format_reference_portfolios_response_success(
+        self, _, response, num_items, expected_value
+    ):
+        succ, err = format_reference_portfolios_response(response)
+        self.assertEqual(num_items, len(succ))
+        self.assertEqual(num_items, len(err))
+
+        [
```
rizwansaeed, line 404: I know this was following the pattern before, but I think this ~can~ should be rewritten as:

```
for index, row in err.iterrows():
    self.assertEqual(expected_value["succ"][index], row[succ.columns[0]])
    self.assertEqual(expected_value["err"][index], row[err.columns[0]])
```

tests/unit/cocoon/test_cocoon_printer.py
```diff
@@ -363,6 +383,33 @@ def test_format_holdings_response_success(
...
+                {"succ": ["ID00002", "ID00002"], "err": ["not found", "not found"]},
+            ),
+            ("empty_response", empty_response_with_full_shape, 0, {}),
+        ]
+    )
+    def test_format_reference_portfolios_response_success(
+        self, _, response, num_items, expected_value
+    ):
+        succ, err = format_reference_portfolios_response(response)
+        self.assertEqual(num_items, len(succ))
```
rizwansaeed, line 401: I think  you can appease sonar by pulling the assertions into a new function, something like

`def assertResponses(self, succ, errr):`

jyaffe:
All of the standard response tests in the CocoonPrinterTests class fail with this assert_responses function

PR #140, opened 2020-07-28
Feature/se 146.seed data optional mappings default

jyaffe:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

When explicitly passing in a mapping to `seed_data` this commit defaults any optional mappings to an empty dict/list if one is not provided. This is a requirement of lpt rather than a requirement of the core API. 
https://finbourne.atlassian.net/browse/SE-146

### DIFF
--- a/lusidtools/cocoon/seed_sample_data.py
+++ b/lusidtools/cocoon/seed_sample_data.py
@@ -67,7 +67,7 @@ def seed_data(
 
         data_frame = getattr(pd, f"read_{supported_files[file_type]}")(transaction_file)
 
-    def check_or_set_default_value(mapping, key, default_value):
+    def check_or_set_default_value(mapping, check_key, default_value):
         """
         This function check whether the mappings variables have the required default values to be uploaded
         via the load_from_data_frame function.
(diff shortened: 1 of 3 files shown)


lusidtools/cocoon/seed_sample_data.py
```diff
@@ -66,6 +67,35 @@ def seed_data(
...
+            an object such as an empty dictionary or list
+
+        Returns
+        -------
+        mapping : dict
+            a file containing mapping of DataFrame headers to LUSID headers.
+
+        """
+
+        for key, value in mapping.items():
```
rizwansaeed, line 93: you can simplify this to

```
if check_key not in mapping:
    value.update({check_key: default_value})
```

jyaffe: yes - you're right. will fix and re-push

lusidtools/cocoon/seed_sample_data.py
```diff
@@ -66,6 +67,35 @@ def seed_data(
...
+            if check_key not in value.keys():
```
rizwansaeed, line 94: for safety/future proof also add a check that value is a dict e.g.
```
if isinstance(value, dict) and check_key not in value.keys():
```

PR #141, opened 2020-07-28
Feature/se 228

slemasne:
# Pull Request Checklist

- [ x] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/docs/CONTRIBUTING.md)
- [ x] Tests pass

# Description of the PR

This PR has new function which has the impact of amending a current transaction type configuration.

The function works as follows:

(1) Query all available transaction types in LUSID
(2) Check if the alias exists
(3) If alias does not already exist, throw a Warning
(4) If alias does exist, delete the current configuration and replace with configuration defined in function
(5) Set the new config

### DIFF
--- a/lusidtools/cocoon/transaction_type_upload.py
+++ b/lusidtools/cocoon/transaction_type_upload.py
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
     )
 
     return response
+
+
+def replace_current_transaction_alias(
+    api_factory, updated_alias, movements, properties=None
+):
+
+    """
+    This function overrides the current transaction types alias with a new configuration
+
+    Parameters
+    ----------
+    api_factory: lusid.utilities.ClientApiFactory
+        The LUSID api factory to use.
+    updated_alias: lusid.models.TransactionConfigurationTypeAlias
+        The alias which we want to update.
+    movements: list[lusid.models.TransactionConfigurationMovementDataRequest]
+        The movements to use for transaction type
+    properties: dict[str, lusid.PerpetualProperty]
+        The custom properties to assign to the transaction type
+
+    Returns
+    -------
+    response : (lusid.models.TransactionSetConfigurationData)
+        The response from setting the transaction type
+    """
+
+    # Call LUSID to get a list of the current transaction types
+
+    current_transaction_types = api_factory.build(
+        lusid.api.SystemConfigurationApi
+    ).list_configuration_transaction_types()
+
+    transaction_configs_list = current_transaction_types.transaction_configs
+
+    # Build the new transaction type you want to upload
+
+    new_transaction_request = models.TransactionConfigurationDataRequest(
+        aliases=[updated_alias], movements=movements, properties=properties,
+    )
+
+    # Delete the current configuration assigned to your replacement alias
+
+    for txn_index, trans_type in enumerate(transaction_configs_list):
+
+        find_new_alias = False
+
+        for alias_index, alias in enumerate(trans_type.aliases):
+
+            if (
+                alias.type == updated_alias.type
+                and alias.transaction_group == updated_alias.transaction_group
+            ):
+
+                del trans_type.aliases[alias_index]
+
+                find_new_alias = True
+
+                break
+
+        else:
+            continue
+
+        # If there are no aliases assigned against the old movement, delete the transaction type configuration from LUSID
+        # We don't want to leave unassigned movements in LUSID
+
+        if len(trans_type.aliases) == 0:
+            del transaction_configs_list[txn_index]
+
+        break
+
+    if find_new_alias is False:
+
+        raise Warning(
+            "The requested alias is not available in LUSID. No updates have been made."
+        )
+
+    transaction_configs_list.append(new_transaction_request)
+
+    set_response = api_factory.build(
+        lusid.api.SystemConfigurationApi
+    ).set_configuration_transaction_types(
+        transaction_set_configuration_data_request=models.TransactionSetConfigurationDataRequest(
+            transaction_config_requests=transaction_configs_list,
+            side_config_requests=current_transaction_types.side_definitions,
+        )
+    )
+
+    return set_response
--- a/tests/integration/cocoon/test_replace_current_transaction_alias.py
+++ b/tests/integration/cocoon/test_replace_current_transaction_alias.py
@@ -1,0 +1,138 @@
+import unittest
+from pathlib import Path
+import lusid
+import lusid.models as models
+from lusidtools.cocoon.utilities import create_scope_id
+from lusidtools.cocoon.transaction_type_upload import replace_current_transaction_alias
+import logging
+
+logger = logging.getLogger()
+new_transaction_type = create_scope_id().replace("-", "")
+
+
+class CocoonTestTransactionTypeUpload(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+
+        secrets_file = Path(__file__).parent.parent.parent.joinpath("secrets.json")
+        cls.api_factory = lusid.utilities.ApiClientFactory(
+            api_secrets_filename=secrets_file
+        )
+
+        cls.system_configuration_api = cls.api_factory.build(
+            lusid.api.SystemConfigurationApi
+        )
+
+        cls.alias = [
+            models.TransactionConfigurationTypeAlias(
+                type=new_transaction_type,
+                description="TESTBUY1",
+                transaction_class="TESTBUY1",
+                transaction_group="SYSTEM1",
+                transaction_roles="AllRoles",
+            )
+        ]
+
+        cls.movements = [
+            models.TransactionConfigurationMovementData(
+                movement_types="StockMovement",
+                side="Side1",
+                direction=1,
+                properties={},
+                mappings=[],
+            ),
+            models.TransactionConfigurationMovementData(
+                movement_types="CashCommitment",
+                side="Side2",
+                direction=1,
+                properties={},
+                mappings=[],
+            ),
+        ]
+
+        cls.create_transaction_response = cls.system_configuration_api.create_configuration_transaction_type(
+            transaction_configuration_data_request=models.TransactionConfigurationDataRequest(
+                aliases=cls.alias, movements=cls.movements
+            )
+        )
+
+    def test_update_current_alias_with_new_movements(self):
+
+        updated_movements = [
+            models.TransactionConfigurationMovementData(
+                movement_types="StockMovement",
+                side="Side1",
+                direction=1,
+                properties={},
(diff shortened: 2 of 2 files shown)


lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,82 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+    """
+    This function overrides the current transaction types alias with a new configuration
+
+    Parameters
+    ----------
+    api_factory : lusid.utilities.ClientApiFactory
+        The LUSID api factory to use
+    alias : lusid.models.TransactionConfigurationTypeAlias
+        An aliases with type and group
+    movements : list[lusid.models.TransactionConfigurationMovementDataRequest]
```
MikeMcGarry, line 76: Great having a docstring! Unfortunately these two parameter names don't match the names in the function. Please update.

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,82 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+                del trans_type.aliases[alias_index]
+
+                find_new_alias = True
+
+        # If there are no aliases assigned against the old movement, delete the transaction type configuration from LUSID
+        # We don't want to leave unassigned movements in LUSID
+
+        if len(trans_type.aliases) == 0:
+
+            del transaction_configs_list[txn_index]
```
MikeMcGarry, line 122: Great catch on this edge case

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,82 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+
```
MikeMcGarry, line 123: As the alias must be unique to the transaction type configuration, once you have found it, you can break the outer loop too. Currently it still iterates over all the other transaction types.

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,82 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+        for alias_index, alias in enumerate(trans_type.aliases):
+
+            if (
+                alias.type == updated_alias.type
+                and alias.transaction_group == updated_alias.transaction_group
+            ):
+
+                del trans_type.aliases[alias_index]
+
+                find_new_alias = True
```
MikeMcGarry, line 115: Now that you've found it you can break this inner loop, no need to iterate over the remaining aliases.

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,82 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+        # If there are no aliases assigned against the old movement, delete the transaction type configuration from LUSID
+        # We don't want to leave unassigned movements in LUSID
+
+        if len(trans_type.aliases) == 0:
+
+            del transaction_configs_list[txn_index]
+
+    if find_new_alias is False:
+
+        raise Warning("The requested alias is not available in LUSID.")
```
MikeMcGarry, line 126: Might be worth adding some additional commentary on what the implications (if any?) of this are.

tests/integration/cocoon/test_replace_current_transaction_alias.py
```diff
@@ -0,0 +1,134 @@
...
+
+        for trans_type in get_transaction_types.transaction_configs:
+            for alias in trans_type.aliases:
+                if alias.type == new_transaction_type:
+                    uploaded_alias.append(trans_type)
+
+        alias_for_test = models.TransactionConfigurationData(
+            aliases=[self.alias], movements=updated_movements, properties={}
+        )
+        self.assertEqual(uploaded_alias[0].to_dict(), alias_for_test.to_dict())
```
MikeMcGarry, line 96: Does this assertion fail if you don't convert them to_dict()? You can see [here](https://github.com/finbourne/lusid-sdk-python-preview/blob/master/sdk/lusid/models/transaction_configuration_data.py) in the __eq__ function -  that when you check that these two classes are equal it will check they are of the same class than to a to_dict conversion anyway.

tests/integration/cocoon/test_replace_current_transaction_alias.py
```diff
@@ -0,0 +1,134 @@
...
+        with self.assertRaises(Warning) as error:
+            replace_current_transaction_alias(
+                api_factory=self.api_factory,
+                updated_alias=updated_alias,
+                movement=updated_movements,
+            )
+
+        self.assertEqual(
+            error.exception.args[0], "The requested alias is not available in LUSID."
+        )
```
MikeMcGarry, line 134: What is the behaviour here? Does it still successfully update the transaction type configuration even if the alias does not exist?

lusidtools/cocoon/transaction_type_upload.py
slemasne: Ah, a copy and paste error from me. That's fixed!

lusidtools/cocoon/transaction_type_upload.py
slemasne: Thanks.
I didn't want to leave movements assigned to empty aliases.

lusidtools/cocoon/transaction_type_upload.py
slemasne: The for-else, that's interesting!
http://psung.blogspot.com/2007/12/for-else-in-python.html
Done!

lusidtools/cocoon/transaction_type_upload.py
slemasne: Done

tests/integration/cocoon/test_replace_current_transaction_alias.py
slemasne: It did fail originally as I was using a **_TransactionConfigurationMovementDataRequest_** rather than a **_TransactionConfigurationMovementData_**.
I updated the model and the eq method gives correct results.

tests/integration/cocoon/test_replace_current_transaction_alias.py
slemasne: The intended behavior - the function request will fail if you try and amend a transaction type which does not exist.
Here is my thinking:
- Transaction types are system wide settings 
- Incorrectly amending them can have bad consequences
- Therefore I would like users to be very explicit when using the function 
- This function is specifically for updating a current transaction type configuration
-  If that transaction type does not exist, I would rather we tell the user (via this error) rather than create the transaction type

lusidtools/cocoon/transaction_type_upload.py
slemasne: Sure, done.

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
     )
 
     return response
+
+
+def replace_current_transaction_alias(
+    api_factory, updated_alias, movement, properties=None
```
MikeMcGarry, line 64: "movement" here "movements" in your docstring

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+
+    Parameters
+    ----------
+    api_factory: lusid.utilities.ClientApiFactory
+        The LUSID api factory to use.
+    updated_alias: lusid.models.TransactionConfigurationTypeAlias
+        The alias which we want to update.
+    movements: list[lusid.models.TransactionConfigurationMovementDataRequest]
+        The movements to use for transaction type
+    properties:
```
MikeMcGarry, line 78: What type should the properties be, what are these can you please add a description and types?

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+                and alias.transaction_group == updated_alias.transaction_group
+            ):
+
+                del trans_type.aliases[alias_index]
+
+                find_new_alias = True
+
+                break
+            else:
+                continue
```
MikeMcGarry, line 119: This else here is not necessary. It will continue anyway without this else clause.

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+        else:
+            continue
+        break
```
MikeMcGarry, line 122: This is nice, so is this saying that if the condition...

``` 
if (
                alias.type == updated_alias.type
                and alias.transaction_group == updated_alias.transaction_group
            ):
```

Is not met at any time in the for loop, then it will continue to the next transaction type, however if it has been met then it will break as there is no need to look at any more transaction types?

lusidtools/cocoon/transaction_type_upload.py
```diff
@@ -58,3 +58,91 @@ def create_transaction_type_configuration(api_factory, alias, movements):
...
+
+        # If there are no aliases assigned against the old movement, delete the transaction type configuration from LUSID
+        # We don't want to leave unassigned movements in LUSID
+
+        if len(trans_type.aliases) == 0:
+
+            del transaction_configs_list[txn_index]
```
MikeMcGarry, line 129: This needs to be moved up after the outside loop else and before the outside loop break, because it is after a break it can never be run and you can see that there is a Sonar error for this.

lusidtools/cocoon/transaction_type_upload.py
slemasne: Thanks. Updated to "movements"

lusidtools/cocoon/transaction_type_upload.py
slemasne: It should be dict[str, lusid.PerpetualProperty].
That's what the TransactionConfigurationDataRequest model takes. 
Updated.

lusidtools/cocoon/transaction_type_upload.py
slemasne: Updated.

lusidtools/cocoon/transaction_type_upload.py
slemasne: Exactly.
Once found, the rest of the loop is redundant. 
There should only be one unique pair of **alias_type and transaction_group** in each LUSID.

lusidtools/cocoon/transaction_type_upload.py
slemasne: Fixed!

PR #143, opened 2020-08-11
Feature/se 275.lpt accept hardcoded int

jyaffe:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Allows for hard-coded integers to be passed as integers into LTP mappings, rather than as strings with the $ constant prefix.


lusidtools/cocoon/utilities.py
```diff
@@ -850,11 +850,15 @@ def handle_nested_default_and_column_mapping(
                 mapping_updated[key] = f"LUSID.{key}"
                 data_frame[mapping_updated[key]] = value[1:]
 
+        elif isinstance(value, int):
```
slemasne, line 853: Nice one!

PR #179, opened 2020-12-09
SE-306 annotates all cocoon tests with Lusid Features

slacki123:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Annotates all cocoon test features with lusid_feature decorators

### DIFF
--- a/tests/integration/cocoon/test_cocoon_failures.py
+++ b/tests/integration/cocoon/test_cocoon_failures.py
@@ -385,9 +385,9 @@ class CocoonTestsFailures(unittest.TestCase):
             ]
         ]
     )
-    def test_load_from_data_frame_failure2(
+    def test_load_from_data_frame_failure_multi_index_dataframe(
         self,
-        test_name,
+        _,
         scope,
         file_name,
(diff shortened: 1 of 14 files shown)


tests/integration/cocoon/test_cocoon_failures.py
```diff
@@ -365,7 +385,7 @@ def test_load_from_data_frame_failure(
             ]
         ]
     )
-    def test_load_from_data_frame_failure(
+    def test_load_from_data_frame_failure2(
```
rizwansaeed, line 388: why was this renamed to `2`? can we have a more descriptive name

rizwansaeed (changes requested):

tests/integration/cocoon/test_cocoon_failures.py
slacki123: @rizwansaeed Turns out that if two methods have the exact same names in the same class, then only one of them is returned by the inspect method. That's the reason for name change

rizwansaeed (approved):

PR #190, opened 2021-01-06
SE-610: Adding support for bond instrument in load_from_data_frame

cscialino:
SE-610: Adding support complex instruments in load_from_data_frame

# Pull Request Checklist

- [ x ] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/docs/CONTRIBUTING.md)
- [ x ] Tests pass

# Description of the PR

- Added support for Bond Instrument definition to ```load_from_data_frame```
- Tests passed

### DIFF
--- a/lusidtools/cocoon/utilities.py
+++ b/lusidtools/cocoon/utilities.py
@@ -269,7 +269,8 @@ def set_attributes_recursive(
             if (key, attribute_type) == ("identifiers", "dict(str, str)"):
                 obj_init_values[key] = {
                     str_key: row[str_value]
-                    for str_key, str_value in mapping[key].items() if not pd.isna(row[str_value])
+                    for str_key, str_value in mapping[key].items()
+                    if not pd.isna(row[str_value])
                 }
             else:
                 obj_init_values[key] = additional_attributes[key]
@@ -290,7 +291,9 @@ def set_attributes_recursive(
                 if "date" in key or "created" in key or "effective_at" in key:
                     obj_init_values[key] = str(DateOrCutLabel(row[mapping[key]]))
                 # Converts to a list element if it is a list field
-                elif 'list' in attribute_type and not isinstance(row[mapping[key]], list):
+                elif "list" in attribute_type and not isinstance(
+                    row[mapping[key]], list
+                ):
                     obj_init_values[key] = [row[mapping[key]]]
                 else:
                     obj_init_values[key] = row[mapping[key]]
@@ -315,7 +318,6 @@ def set_attributes_recursive(
 
             obj_init_values[key] = [value] if nested_type == "list" else value
 
-
     """
     If all attributes are None propagate None rather than a model filled with Nones. For example if a CorporateActionSourceId
     has no scope or code return build a model with CorporateActionSourceId = None rather than CorporateActionSourceId = 
@@ -327,7 +329,7 @@ def set_attributes_recursive(
 
     # Create an instance of and populate the model object
     instance = model_object(**obj_init_values)
-    
+
     # Support for polymorphism, we can identify these `abstract` classes by the existence of the below
     if getattr(instance, "discriminator"):
         discriminator = getattr(instance, getattr(instance, "discriminator"))
@@ -335,9 +337,7 @@ def set_attributes_recursive(
         actual_class = model_object.discriminator_value_class_map[discriminator]
 
         return set_attributes_recursive(
-            model_object=getattr(lusid.models, actual_class),
-            mapping=mapping,
-            row=row,
+            model_object=getattr(lusid.models, actual_class), mapping=mapping, row=row,
         )
 
     return instance
--- a/lusidtools/cocoon/utilities.py
+++ b/lusidtools/cocoon/utilities.py
@@ -260,9 +260,20 @@ def set_attributes_recursive(
     # For each of the attributes to populate
     for key in list(populate_attributes):
 
+        # Get the attribute type
+        attribute_type = obj_attr[key]
+
         # If it is an additional attribute, populate it with the provided values and move to the next attribute
         if key in list(additional_attributes.keys()):
-            obj_init_values[key] = additional_attributes[key]
+            # Handle identifiers provided within instrument definition (e.g. 'Bond', 'Future', etc.)
+            if (key, attribute_type) == ("identifiers", "dict(str, str)"):
+                obj_init_values[key] = {
+                    str_key: row[str_value]
+                    for str_key, str_value in mapping[key].items() if not pd.isna(row[str_value])
+                }
+            else:
+                obj_init_values[key] = additional_attributes[key]
+
             continue
 
         # This block keeps track of the number of missing (non-additional) attributes
@@ -271,9 +282,6 @@ def set_attributes_recursive(
             if mapping[key] is None:
                 none_count += 1
 
-        # Get the attribute type
-        attribute_type = obj_attr[key]
-
         # If this is the last object and there is no more nesting set the value from the row
         if not isinstance(mapping[key], dict):
             # If this exists in the mapping with a value and there is a value in the row for it
@@ -281,6 +289,9 @@ def set_attributes_recursive(
                 # Converts to a date if it is a date field
                 if "date" in key or "created" in key or "effective_at" in key:
                     obj_init_values[key] = str(DateOrCutLabel(row[mapping[key]]))
+                # Converts to a list element if it is a list field
+                elif 'list' in attribute_type and not isinstance(row[mapping[key]], list):
+                    obj_init_values[key] = [row[mapping[key]]]
                 else:
                     obj_init_values[key] = row[mapping[key]]
             elif obj_attr_required_map[key] == "required":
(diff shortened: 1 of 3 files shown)


tests/unit/cocoon/test_cocoon_complex_instrument.py
```diff
@@ -0,0 +1,204 @@
...
+
+        combined_successes = {
+            correlation_id: instrument
+            for response in responses["instruments"]["success"]
+            for correlation_id, instrument in response.values.items()
+        }
+
+        self.assertTrue(
+            expr=all(
+                sorted(combined_successes[key].properties, key=lambda p: p.key)
```
MikeMcGarry, line 196: Does this test illustrate that your change is working, it looks to only check that all the property keys are the same between the test and the expected outcome?

tests/unit/cocoon/test_cocoon_complex_instrument.py
```diff
@@ -0,0 +1,204 @@
...
+                {"Isin": "ISIN", "ClientInternal": "Client Internal"},
+                [],
+                "TestPropertiesScope1",
+            ],
+            [
+                "A standard successful load of a complex instrument using a full bond definition",
+                "TestScope1",
+                "data/global-fund-fixed-income-master.csv",
+                {
+                    "name": "Name",
```
MikeMcGarry, line 103: I don't see where you would ever have a dictionary which has `instrument_type`  as a key based on this mapping here?

lusidtools/cocoon/utilities.py
```diff
@@ -816,9 +817,28 @@ def handle_nested_default_and_column_mapping(
 
         # If the value of the mapping is a dictionary
         if isinstance(value, dict):
+            nested_key = key
+
+            # If an 'instrument_type' is specified within mapping, expand nested definition mapping
+            if "instrument_type" in mapping.keys():
```
MikeMcGarry, line 823: Can we make this agnostic to instruments?

lusidtools/cocoon/utilities.py
```diff
@@ -816,9 +817,28 @@ def handle_nested_default_and_column_mapping(
...
+                    models.LusidInstrument.discriminator_value_class_map
+                )
+                if mapping_object not in valid_instrument_types:
+                    raise TypeError(
+                        f"""The specified instrument type {mapping_object} is invalid.
+                        Please set instrument type to one of: {valid_instrument_types.keys()}"""
+                    )
+                else:
+                    for key, value in mapping[nested_key].items():
+                        mapping_updated[key] = value
```
MikeMcGarry, line 838: Even if the mapping updated how would this get used later on? There is no field called `dom_ccy` as per your test the mapping would need to be `definition.dom_ccy`.

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
+            # Handle identifiers provided within instrument definition (e.g. 'Bond', 'Future', etc.)
+            if (key, attribute_type) == ("identifiers", "dict(str, str)"):
+                obj_init_values[key] = {
+                    str_key: row[str_value["value"]]
+                    for str_key, str_value in mapping[key].items()
+                }
+                continue
+            else:
+                obj_init_values[key] = additional_attributes[key]
+                continue
```
MikeMcGarry, line 277: You will continue here no matter what, so you can put continue just once outside the if/else on line 277 and remove it on line 274.

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
 
+        # Get the attribute type
+        attribute_type = obj_attr[key]
+
         # If it is an additional attribute, populate it with the provided values and move to the next attribute
         if key in list(additional_attributes.keys()):
-            obj_init_values[key] = additional_attributes[key]
-            continue
+            # Handle identifiers provided within instrument definition (e.g. 'Bond', 'Future', etc.)
+            if (key, attribute_type) == ("identifiers", "dict(str, str)"):
```
MikeMcGarry, line 269: Ahhh very nice, great pickup that we are considering `identifiers` to be an exception to how we process most attributes,  clearly with these `LusidInstruments` that exception is broken. This seems to be a sensible way of handling the exception to the exception.

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
     for key in list(populate_attributes):
 
+        # Get the attribute type
+        attribute_type = obj_attr[key]
+
         # If it is an additional attribute, populate it with the provided values and move to the next attribute
         if key in list(additional_attributes.keys()):
-            obj_init_values[key] = additional_attributes[key]
-            continue
+            # Handle identifiers provided within instrument definition (e.g. 'Bond', 'Future', etc.)
```
MikeMcGarry, line 268: Love this comment, very useful thank you!

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
+            if (key, attribute_type) == ("identifiers", "dict(str, str)"):
+                obj_init_values[key] = {
+                    str_key: row[str_value["value"]]
```
MikeMcGarry, line 271: Nice dictionary comprehension!

MikeMcGarry, line 271: Why do we take `str_value["value"]`? My understanding is that `str_value` is a string, wouldn't calling `str_value["value"]` result in an Exception being thrown?

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
+                    for str_key, str_value in mapping[key].items()
```
MikeMcGarry, line 272: Do we need to add `if not pd.isna(row[mapping[key]])` to ensure that we don't attempt to get a value from the DataFrame when it is `NaN` or `None`?

lusidtools/cocoon/utilities.py
```diff
@@ -260,27 +260,40 @@ def set_attributes_recursive(
...
             # If this exists in the mapping with a value and there is a value in the row for it
             if mapping[key] is not None and not pd.isna(row[mapping[key]]):
                 # Converts to a date if it is a date field
                 if "date" in key or "created" in key or "effective_at" in key:
                     obj_init_values[key] = str(DateOrCutLabel(row[mapping[key]]))
+                # Converts to a list element if it is a list field
+                elif "list" in attribute_type and not isinstance(
+                    row[mapping[key]], list
+                ):
+                    obj_init_values[key] = list(row[mapping[key]].split())
```
MikeMcGarry, line 296: Does `obj_init_values[key] = [row[mapping[key]]]` work?  That is what we use on line 318 below and it is simpler than splitting.

lusidtools/cocoon/utilities.py
```diff
@@ -314,7 +327,24 @@ def set_attributes_recursive(
...
     # Create an instance of and populate the model object
-    return model_object(**obj_init_values)
+    instance = model_object(**obj_init_values)
+
+    # Support for polymorphism, we can identify these `abstract` classes by the existence of the below
+    if getattr(instance, "discriminator"):
+        discriminator = getattr(instance, getattr(instance, "discriminator"))
+        try:
+            actual_class = model_object.discriminator_value_class_map[discriminator]
+        except TypeError:
```
MikeMcGarry, line 337: This will likely be a `KeyError` rather than a `TypeError`.

lusidtools/cocoon/utilities.py
```diff
@@ -314,7 +327,24 @@ def set_attributes_recursive(
...
+            print(
+                f"Invalid discriminator object {discriminator} was passed. "
+                f"Please ensure is one of: {model_object.discriminator_value_class_map}"
+            )
```
MikeMcGarry, line 341: Great checking, can we please use `logging.error` instead of `print`?

tests/unit/cocoon/test_cocoon_complex_instrument.py
```diff
@@ -0,0 +1,176 @@
...
+from parameterized import parameterized
+import lusid
+import lusid.models as models
+from lusid.utilities import ApiClientFactory
+from lusidtools import logger
+from datetime import datetime
+import pytz
+
+
+class ComplexInstrumentTests(unittest.TestCase):
```
MikeMcGarry, line 15: Great test, can you please add another test to test the failure case where the discriminator isn't valid?

MikeMcGarry (changes requested):
Looking great, fantastic job!

A couple of places I think we can make some slight optimisations which I have made comments for.

On line 271 I need to understand how ` row[str_value["value"]]` works.

And finally can we please add a test for the failure case in which you provide an invalid instrument type e.g. "UnsupportedInstrument" to ensure we understand what the behaviour will be and assert that the error raised is as we expect.

lusidtools/cocoon/utilities.py
cscialino: Certainly!

lusidtools/cocoon/utilities.py
cscialino: Yes you are right. In fact something to add a test for, and I think we'll want to ensure it opens an empty dict() when it's NaN. Otherwise, one is required to map the identifier a second time, which is a mistake that people will surely make given the variable is set as nullable.

MikeMcGarry: Yes, great point about a dictionary if all are NaN! I wonder if that would be the behaviour anyway if all are NaN or not? Like you say best way to see is via a test!

PR #191, opened 2021-01-08
Feature/se 672

MikeMcGarry:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


lusidtools/cocoon/utilities.py
```diff
@@ -282,8 +283,10 @@ def set_attributes_recursive(
                     obj_init_values[key] = str(DateOrCutLabel(row[mapping[key]]))
                 else:
                     obj_init_values[key] = row[mapping[key]]
-            elif mapping[key]:
+            elif obj_attr_required_map[key] == "required":
```
rizwansaeed, line 286: this might need a check to validate the key is in the dict

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -648,6 +648,12 @@ def test_load_instrument_properties_with_missing_instruments(self):
         ]
     )
     def test_load_instrument_lookthrough(self, _, df, mapping):
+
+        print(_)
```
rizwansaeed, line 652: remove?

lusidtools/cocoon/utilities.py
MikeMcGarry: The keys for `model_object.required_map` are the same as for `model_object.openapi_types` so checking for the existence of `key` in the dictionary this is not necessary as the only keys you will see at this point must be in  `model_object.openapi_types` .

rizwansaeed: `key` looks like its coming from `provided_attributes` https://github.com/finbourne/lusid-python-tools/blob/2739a2faa073423c773d5213182d58aa16f84655/lusidtools/cocoon/utilities.py#L251 which suggests it might be user supplied and needs validating?

MikeMcGarry: We take an intersection two lines below which ensures we don't have a key which isn't populatable.

`populate_attributes = provided_attributes.intersection(available_attributes)`

rizwansaeed (approved):

PR #192, opened 2021-01-20
SE-709: Update Command line app reader

thomasdotred:
Remove expected index column for LPT command line apps

# Pull Request Checklist

- [X] Tests pass

# Description of the PR

passing in a csv without an index column will result in the first column being ignored (by default pandas expects an index column)

This is fixed by setting `index_col=None` to the read_csv command

also added support for excel files


rizwansaeed (approved):

tests/unit/apps/test_command_line_utilities.py
```diff
@@ -71,6 +71,34 @@ def test_get_instruments_data(self, _, args_list):
             self.assertEqual(data.tail(1).values[0][0], "Whitebread_LondonStockEx_WTB")
         else:
             self.assertEqual(data.tail(1).values[0][0], "USTreasury_6.875_2025")
+
+    @parameterized.expand(
```
rizwansaeed, line 75: don't need to parameterise this

tests/unit/apps/test_command_line_utilities.py
```diff
@@ -71,6 +71,34 @@ def test_get_instruments_data(self, _, args_list):
...
+                    "delimiter": None,
+                    "scale_quotes": False,
+                    "line_terminator": None,
+                },
+            ]
+        ]
+    )
+    def test_get_instruments_data_excel(self, _, args_list):
+        data = load_data_to_df_and_detect_delimiter(args_list)
+        if args_list["num_header"] == 0:
```
rizwansaeed, line 94: avoid having conditional logic in test asserts, suggests they should be separate tests

rizwansaeed (changes requested):

PR #195, opened 2021-03-01
SE-852 Added support for providing a scope for the sub-holding-keys w…

MikeMcGarry:
…hich may be different to both the scope and the properties scope. This is an optional keyword parameter on load_from_data_frame

# Pull Request Checklist

- [ ] Tests pass

# Description of the PR


lusidtools/cocoon/cocoon.py
```diff
@@ -661,7 +664,7 @@ def _convert_batch_to_models(
         ):
             sub_holding_keys_row = cocoon.properties.create_property_values(
                 row=row,
-                scope=properties_scope,
+                scope=sub_holding_keys_scope,
```
kkeenan-finbourne, line 667: ~~Does this possibly introduce a breaking a change? Thinking of a scenario of anyone currently using LPT who is assuming that SHKs scope are the property scopes? In which case should this check for `sub_holdings_keys_scope` being null or empty and revert to `property_scope`? Or are we looking for a clean change and do we need to document somewhere? wiki maybe?~~

kkeenan-finbourne: Ignore above I missed that you already do that below in load_from_data_frame

kkeenan-finbourne (approved):

thomasdotred (approved):

PR #203, opened 2021-04-26
SE-1134: return_unmatched_identifiers flag in load_from_data_frame

jyaffe:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

This PR introduces an optional flag in the `load_from_data_frame()` cocoon method. 

The flag defaults to `False` but when `True` is passed for `file_type="holding"` or `file_type="transaction"`, an additional field (`unmatched_identifiers`) will be included in the response.

The `unmatched_identifiers` field will provide the user with a list of instruments that were part of the upload but were not able to be resolved within LUSID. In other words, the holdings or transactions that were uploaded will show an `instrumentUid` with the value `LUID_ZZZZZZZZ` which indicates that the instrument has not yet been created in LUSID. 

The user will be able to use this additional response to identify which parts of the holding or transaction upload require further instrument details. 

For holdings, the response in `unmatched_identifiers` will consist of a list of `instrument_identifier` dictionaries. 

For transactions, the response in `unmatched_identifiers` will consist of a list of objects with the structure `{ transaction_id: instrument_identifiers}`

TODO:
- [x] The final two test cases in `test_cocoon_holdings.test_load_from_data_frame_holdings_success_with_unmatched_identifiers()` will occasionally fail due to the variable ordering of the response. This needs to be fixed before the PR can be merged
- [x] There may be further edge-cases that we wish to test


lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+
+    Returns
+    -------
+    A filtered list of unmatched_transactions.
+    """
+    # Create a list var to hold the filtered transactions to be returned
+    filtered_unmatched_transactions = []
+
+    # Create a unique list of transaction_ids from the dataframe
+    valid_txids = data_frame[mapping_required.get("transaction_id")].unique()
```
rizwansaeed, line 1150: make this a set so the `in` check is O(1) lookup

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+
+    # Iterate through the transactions and if the transaction_id is in the dataframe, add it to the list to be returned
+    for unmatched_transaction in unmatched_transactions:
```
rizwansaeed, line 1153: if `unmatched_transactions` is a unique list, you can use `set.difference`

lusidtools/cocoon/cocoon.py
thomasdotred: for dict.get(): Average case O(1); Amortised analysis case O(N)

(https://www.geeksforgeeks.org/complexity-cheat-sheet-for-python-operations/)

lusidtools/cocoon/cocoon.py
thomasdotred: set operations should be significantly faster for operations that check the values within them (as opposed to iteration over its items where lists are faster)
This is a super important when we have portfolios with > 100,000s of txns

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+    # Create empty list to hold instrument identifiers that have not been resolved
+    unmatched_instruments = []
+
+    # For each holding adjustment in the upload, check whether any contained unresolved instruments and append to list
+    for code_tuple in code_tuples:
+        unmatched_instruments.extend(
+            return_unmatched_instruments_from_holdings(
+                api_factory=api_factory, scope=scope, code_tuple=code_tuple,
+            )
+        )
```
thomasdotred, line 1198: Same point of using a list comprehension as I made above

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+    transactions_api = api_factory.build(lusid.api.TransactionPortfoliosApi)
+    unmatched_instruments = []
+
+    response = transactions_api.get_holdings_adjustment(
+        scope=scope, code=code_tuple[0], effective_at=str(DateOrCutLabel(code_tuple[1]))
+    )
+
+    for adjustment in response.adjustments:
+        if adjustment.instrument_uid == "LUID_ZZZZZZZZ":
+            unmatched_instruments.append(adjustment.instrument_identifiers)
```
thomasdotred, line 1236: List comprehension should significantly speed this up (list comp vs append)

```
unmatched_instruments = [
    adjustment.instrument_identifiers
    for adjustment in response.adjustments if adjustment.instrument_uid == "LUID_ZZZZZZZZ"
]
```

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+            mapping_required=mapping_required,
+            file_type=file_type,
+            sync_batches=sync_batches,
+        )
+
+    return returned_response
+
+
+@checkargs
+def unmatched_identifiers(
```
thomasdotred, line 927: Have you considered importing these additional functions from the cocoon/utilities.py, rather than having them defined here? This would:
- make the cocoon.py file more readable
- encourages you to add unittests for some of these

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+
+    # With the upload dataframe, filter out any transactions from the list that were not part of this upload and return
+    return filter_unmatched_transactions(
+        data_frame=data_frame,
+        mapping_required=mapping_required,
+        unmatched_transactions=unmatched_transactions,
+    )
+
+
+def _extract_unique_portfolio_codes(sync_batches: list):
```
thomasdotred, line 1030: We could add unitttests for these functions

thomasdotred (changes requested):
Hey great work, you have good test coverage 👍  

I have a few points to add, mainly around:
- list comps and sets where possible
- test case where a portfolio has existing zzzz but was not included in the upsert  / set request
- moving some of the functions to utilities.py and adding unit tests

I'll give you a call 👍

lusidtools/cocoon/cocoon.py
jyaffe: I think i want Set.intersection() here, rather than difference. The values that are in the returned list and the acceptable list.

jyaffe: Although as unmatched_transactions is a list of dictionaries, it will not be converted into a set. I think I can make the valid_txids a set but the for loop will have to stay.

lusidtools/cocoon/cocoon.py
```diff
@@ -901,11 +904,342 @@ async def _construct_batches(
...
+    # Create empty list to hold transaction ids and instruments
+    unmatched_transactions = []
+
+    # For each portfolio, request the unmatched transactions from LUSID and append to the instantiated list
+    for portfolio_code in portfolio_codes:
+        unmatched_transactions.extend(
+            return_unmatched_transactions(
+                api_factory=api_factory, scope=scope, code=portfolio_code,
+            )
+        )
```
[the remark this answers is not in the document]
jyaffe: As discussed, return_unmatched_transactions returns a list of objects. If we use a list comprehension we would end up with a list of lists, which would need flattening. This current logic is not more complex than that.

lusidtools/cocoon/cocoon.py
jyaffe: Similar to previous comments, return_unmatched_instruments_from_holdings returns a list of objects. If we use a list comprehension we would end up with a list of lists, which would need flattening. This current logic is not more complex than that.

lusidtools/cocoon/cocoon.py
jyaffe: List comprehension works well here 👍 thanks

lusidtools/cocoon/cocoon.py
jyaffe: I have moved them to utilities but unit tests are needed now

lusidtools/cocoon/cocoon.py
jyaffe: added

lusidtools/cocoon/cocoon.py
jyaffe: tests added

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -19,739 +19,690 @@ def setUpClass(cls) -> None:
...
+    #
+    #     # Assert that by default no unmatched_identifiers are returned in the response
+    #     self.assertFalse(responses["holdings"].get("unmatched_identifiers", False))
+    #
+    #     self.assertTrue(
+    #         expr=all(
+    #             isinstance(success_response.version, expected_outcome)
+    #             for success_response in responses["holdings"]["success"]
+    #         )
+    #     )
```
thomasdotred, line 692: include these tests 👍

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -680,9 +680,170 @@ def test_load_from_data_frame_holdings_success(
...
+                [],
+            ],
+            [
+                "Standard successful load with one unmatched instrument",
+                "data/holdings-example-unique-date-one-unmatched-instrument.csv",
+                True,
+                [
+                    {
+                        "Instrument/default/Sedol": "BFAKE712",
+                        "Instrument/default/Isin": "USFAHHDAR1014",
```
rizwansaeed, line 708: is this a fake ISIN too? it should only return a LUID_ZZZZZZ if none of the identifiers match

you might actually need a test that covers this e.g. valid ISIN, invalid SEDOL should not return ZZZ and unmatched should be empty

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -680,9 +680,170 @@ def test_load_from_data_frame_holdings_success(
...
         self.assertTrue(
             expr=all(
-                isinstance(succcess_response.version, expected_outcome)
-                for succcess_response in responses["holdings"]["success"]
+                isinstance(success_response.version, expected_outcome)
+                for success_response in responses["holdings"]["success"]
             )
         )
+
+    @parameterized.expand(
```
rizwansaeed, line 693: I think we need a test for when `return_unmatched_identifiers` is `False` but there are unmatched identifiers

tests/integration/cocoon/test_cocoon_holdings.py
jyaffe: Good point. Test added, although the code handles this correctly already.

tests/integration/cocoon/test_cocoon_holdings.py
jyaffe: Added

thomasdotred (approved):

.github/workflows/run-tests-on-pull.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     strategy: 
       max-parallel: 1
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
```
rizwansaeed, line 44: can you put his back, if the tests take more than 30mins we've got an issue

lusidtools/cocoon/cocoon.py
```diff
@@ -1191,15 +1188,45 @@ def return_unmatched_instruments_from_holdings(
     )
 
     # Create a list of instrument identifiers (with LUID_ZZZZZZZZ) from the response
-    unmatched_instruments = [
-        adjustment.instrument_identifiers
-        for adjustment in response.adjustments
-        if adjustment.instrument_uid == "LUID_ZZZZZZZZ"
-    ]
+    unmatched_instruments = []
+    for adjustment in response.adjustments:
```
rizwansaeed, line 1192: can this be list comprehension?

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -725,66 +746,146 @@ def test_load_from_data_frame_holdings_success(
             [
```
rizwansaeed, line 746: need to add the `lusid_feature`

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -725,66 +746,146 @@ def test_load_from_data_frame_holdings_success(
...
-                    },
-                    {
-                        "Instrument/default/Sedol": "BFAKE712",
-                        "Instrument/default/Isin": "USFAHHDAR1014",
-                    },
-                ],
+                holding_instrument_identifiers(fake1=True, fake2=False),
+            ],
+            [
+                "One portfolio, one sub-holding-key, two dates, two instruments - AdjustHolding",
```
rizwansaeed, line 860: are these all SetHoldings? I think we should add explicitly whether each one is Set or Adjust

jyaffe: Good point

lusidtools/cocoon/cocoon.py
jyaffe: Leaving this for now

tests/integration/cocoon/test_cocoon_holdings.py
jyaffe: Added the features for new tests and found some existing inconsistencies that I've cleaned up

rizwansaeed:
can you fix up the code smells for using constants

rizwansaeed (approved):

PR #204, opened 2021-05-06
SE-1201: Added flag for extended errors in cocoon printer

jyaffe:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

This PR will add a flag to the cocoon printer formatting methods to enable extended error message reporting. 

By default the behavior is the same as before. With the new flag being passed, the errors in the response formatting will include the RequestIds and the ErrorDetails to help the user identify potential issues faster.

### DIFF
--- a/lusidtools/cocoon/cocoon_printer.py
+++ b/lusidtools/cocoon/cocoon_printer.py
@@ -34,7 +34,7 @@ def check_dict_for_required_keys(
             )
 
 
-def get_errors_from_response(list_of_API_exceptions: list):
+def get_errors_from_response(list_of_API_exceptions: list, extended_error_details: bool = False):
     """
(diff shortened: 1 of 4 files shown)


jyaffe:
NB: I want https://github.com/finbourne/lusid-python-tools/pull/203 to merge before this one.

thomasdotred (approved):
My only comment would be that maybe we shouldn't have the extended_error_details behaviour within the get_errors_from_response function. It is too high level behaviour for and would be better to gave this return all the error details, then have the printers themselves decide whether or not to return the extended details

However, I think it would make sense to leave it here until we merge all printers into a single function, otherwise we will end up repeating this logic across all printers 

(https://finbourne.atlassian.net/browse/SE-591)

PR #206, opened 2021-05-26
SE-1284: Return full response

Omar-Oz:
PR to return full responses for Holdings and transactions.


jyaffe (commented):
Thanks for this Omar. I think it would be worth a chat next week to go through this and discuss the changes and how to work with the parameterised testing.

PR #207, opened 2021-05-29
Feature/se 1284.return full response

jyaffe:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

A recently added flag within `load_from_data_frame()` called `return_unmatched_identifiers` was useful in showing which instrument identifiers did not resolve when uploading transactions or holdings, but there was not always enough information to be able to quickly identify which transactions or holdings were affected. 

This change renames that feature flag to `return_unmatched_items` and updates the logic so that the entire transaction or holding object is returned when the instrument does not successfully resolve in LUSID. This will give all users all the information they require to quickly identify which items in their upload did not resolve, as well as providing them with helpful associated metadata that they can use to improve their workflow to correct the identifiers.

### DIFF
--- a/lusidtools/cocoon/cocoon.py
+++ b/lusidtools/cocoon/cocoon.py
@@ -1072,7 +1072,7 @@ def return_unmatched_transactions(
 
         unmatched_transactions.extend(
             [
-                {response.transaction_id: response.instrument_identifiers}
+                response
                 for response in response.values
             ]
         )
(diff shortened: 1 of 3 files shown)


rizwansaeed (approved):

rizwansaeed:
its helpful to make refactoring in a separate commit so the actual change doesn't get lost in the noise of lots of changes

PR #214, opened 2021-06-18
SE-1323: Handle getholding lusid exception within LPT

jyaffe:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

If a load_from_data_frame encounters an error with a holdings upload, the return_unmatched_items logic (if a True parameter is provided) may obscure the load error by raising it's own exception. The return_unmatched_items error should be handled gracefully and the main lusid error should returned as normal to the user so that they can debug the primary issue.


slemasne (approved):
Approving however make sure the tests run before merging.

PR #215, opened 2021-06-24
SE-1437: Do not validate upload if any errors

jyaffe:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

If any errors are present in an upload using `load_from_data_frame`, the upload validation logic in `return_unmatched_items` will not be run, even if the flag is passed. All errors will be handled and returned to the user.

### DIFF
--- a/tests/integration/cocoon/test_cocoon_holdings.py
+++ b/tests/integration/cocoon/test_cocoon_holdings.py
@@ -1,4 +1,5 @@
 import unittest
+import json
 from pathlib import Path
 import pandas as pd
 from datetime import datetime
@@ -970,7 +971,9 @@ class CocoonTestsHoldings(unittest.TestCase):
         Test that holdings can be loaded successfully
 
         :param str file_name: The name of the test data file
-        :sub_holding_keys
+        :param list[str] sub_holding_keys: The column names for any sub-holding-keys
+        :param bool holdings_adjustment_only: Whether to use the set or adjust holding api
+        :param int expected_holdings_in_response: The number of expected holdings in the response
         :param list[instrument_identifiers] expected_unmatched_identifiers: A list of expected instrument_identifiers
 
         :return: None
@@ -1060,7 +1063,7 @@ class CocoonTestsHoldings(unittest.TestCase):
                 scope=scope, code=portfolio.id.code
             )
 
-    @lusid_feature("T3-42",)
+    @lusid_feature("T3-42")
     def test_return_unmatched_holding_handles_empty_holdings_exception_in_lusid(self):
         """
         Test that the get holdings call to LUSID handles cases where there are no holdings present for a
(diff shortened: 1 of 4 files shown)


tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -970,7 +971,9 @@ def test_load_from_data_frame_holdings_success_with_unmatched_items(
         Test that holdings can be loaded successfully
 
         :param str file_name: The name of the test data file
-        :sub_holding_keys
+        :param list[str] sub_holding_keys: The column names for any sub-holding-keys
+        :param bool holdings_adjustment_only: Whether to use the set or adjust holding api
+        :param int expected_holdings_in_response: The number of expected holdings in the response
         :param list[instrument_identifiers] expected_unmatched_identifiers: A list of expected instrument_identifiers
 
         :return: None
```
thomasdotred, line 979: Ideally, we would be consistently using the same numpy docstring format across all of LPT. This could probably be left for a `good-first-issue` task

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -1110,3 +1113,114 @@ def test_return_unmatched_holding_handles_empty_holdings_exception_in_lusid(self
...
+        Test that a failed holding upload is handled gracefully by the whole load_from_data_frame function.
+        If any upload causes an error the validation logic will not run and all the errors will be returned
+        by load_from_data_frame.
+
+        :param str file_name: The name of the test data file
+        :param list[str] sub_holding_keys: The column names for any sub-holding-keys
+        :param str error_name: The name of the ApiException that is hit by this error
+        :param bool skip_portfolio: Flag whether to skip the creation of a test portfolio
+
+        :return: None
```
thomasdotred, line 1149: Again, would be ideal to use numpy docstring format

tests/integration/cocoon/test_cocoon_holdings.py
```diff
@@ -1110,3 +1113,114 @@ def test_return_unmatched_holding_handles_empty_holdings_exception_in_lusid(self
...
+                    "code": "FundCode",
+                    "display_name": "FundCode",
+                    "base_currency": "Local Currency Code",
+                },
+                file_type="portfolio",
+                mapping_optional={"created": "Created Date"},
+            )
+
+            # Assert that all portfolios were created without any issues
+            self.assertEqual(len(portfolio_response.get("portfolios").get("errors")), 0)
```
thomasdotred, line 1189: In order to be truly independent, we would have different portfolios for each parametrised test case. Alternatively, we could initialise and destroy the portfolio within the setup() and teardown() functions.

tests/integration/cocoon/test_cocoon_holdings.py
jyaffe: Yeah I have noticed this. Agree it is a good first task. Raised [SE-1447](https://finbourne.atlassian.net/browse/SE-1447)

tests/integration/cocoon/test_cocoon_holdings.py
jyaffe: The portfolios are unique for each test case, insofar as the portfolio is created based on the portfolio present in the test data, which is unique and provided by the params. The reason for the skipping portfolio logic here is so that test cases which do not want a portfolio to exist, are not forced to create one. If we want to go the way of setup and teardown functions we could, but it may get a bit complicated due to the fact that the portfolio(s) are different for each test, especially with other tests in this class that are not in this MR.

thomasdotred (approved):

PR #217, opened 2021-07-12
Add error message if config and secrets are missing

mneedham:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR


thomasdotred:
Looks good but we need to deal with the sonar issues here. It looks like an exception to this security vulnerability rule, but I am not sure what the best way around this. Lets raise internally

mneedham:
> Looks good but we need to deal with the sonar issues here. It looks like an exception to this security vulnerability rule, but I am not sure what the best way around this. Lets raise internally

Have updated the PR so that it reads the password from an environment variable and references a non existent file that isn't in the `/tmp` directory

thomasdotred (approved):

PR #218, opened 2021-08-03
SE-147: Flag to include all fields in formatted instruments response

mneedham:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR


tests/integration/cocoon/test_cocoon_format_instruments.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def test_all_fields(self):
+        scope = f"TestScope-{uuid.uuid4()}"
+
+        portfolios = pd.DataFrame(
+            {"code": "Child", "name": "Child Portfolio", "currency": "GBP", "date": "01/01/2020"},
+            {"code": "Parent", "name": "Parent Portfolio", "currency": "GBP", "date": "01/01/2020"}
+        )
+
+        cocoon.cocoon.load_from_data_frame(
```
thomasdotred, line 49: Can we check here that the portfolios do get upserted here? If this request fails then the validation below will fail, so it would be better to raise an error here in this case

tests/integration/cocoon/test_cocoon_format_instruments.py
```diff
@@ -0,0 +1,104 @@
...
+        cls.mapping_optional = {
+            "look_through_portfolio_id.scope": "lookthrough_scope",
+            "look_through_portfolio_id.code": "lookthrough_code",
+        }
+        cls.identifier_mapping = {
+            "Instrument/default/Figi": "figi",
+            "Instrument/default/ClientInternal": "client_internal",
+        }
+
+    def test_all_fields(self):
```
thomasdotred, line 31: Can we rename to `test_format_instrument_response_test_all_fields(self)`

tests/integration/cocoon/test_cocoon_format_instruments.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+        response = cocoon.format_instruments_response(responses, data_entity_details=True)
+        success_df = response[0]
+
+        self.assertEqual(success_df.shape, (1, 14))
+        self.assertEqual(scope, success_df["lookthrough_portfolio.scope"].values[0])
+        self.assertEqual("Child", success_df["lookthrough_portfolio.code"].values[0])
+
+    def test_identifier(self):
```
thomasdotred, line 95: Would be duplicating the unittests in tests/unit/cocoon/test_cocoon_printer.py?

tests/integration/cocoon/test_cocoon_format_instruments.py
```diff
@@ -0,0 +1,104 @@
...
+        responses = cocoon.cocoon.load_from_data_frame(
+            api_factory=self.api_factory,
+            scope=scope,
+            data_frame=instruments,
+            mapping_required=self.mapping_required,
+            mapping_optional=self.mapping_optional,
+            file_type="instruments",
+            identifier_mapping=self.identifier_mapping,
+            instrument_name_enrichment=True,
+        )
```
thomasdotred: Just thinking, do we need to make actual calls to LUSID to test this functionality? Could we just mock the response from LUSID and put it under unittests rather than integration tests?

It could reduce dependency on the LUSID API and would have less of an impact on the overall duration of tests

tests/integration/cocoon/test_cocoon_format_instruments.py
mneedham: Yeh, you're right. Let me add an extra test to `tests/unit/cocoon/test_cocoon_printer.py` instead.

tests/integration/cocoon/test_cocoon_format_instruments.py
mneedham: Yeh that makes sense, have updated the PR

thomasdotred (approved):

PR #219, opened 2021-08-04
CocoonPrinter that encapsulates formatting of different responses

mneedham:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

At the moment it delegates to the functions in `cocoon_printer.py` so that we don't break anything that's already using those.   I've also duplicated the tests so that it tests both ways of formatting responses.


mneedham:
@MikeMcGarry if you have time, could you take a look at this PR?

PR #229, opened 2021-08-10
SE-650: Switch xlrd to openpyxl

mneedham:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ PyYAML==5.4
 requests==2.25.0
 tqdm==4.52.0
 openpyxl==3.0.7
+xlrd==1.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,4 @@ pandas==1.1.4
 PyYAML==5.4
 requests==2.25.0
 tqdm==4.52.0
-xlrd==1.2.0
+openpyxl==3.0.7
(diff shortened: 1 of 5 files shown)


rizwansaeed:
does this mean all current usages need to specify `engine="openpyxl"` or is the default still `xlrd` and specifying the engine allows you to take advantage of the update?

mneedham:
Yeh the default does still seem to be `xlrd`:

```
    def __init__(self, path_or_buffer, engine=None):
        if engine is None:
            engine = "xlrd"
            if isinstance(path_or_buffer, (BufferedIOBase, RawIOBase)):
                if _is_ods_stream(path_or_buffer):
                    engine = "odf"
            else:
                ext = os.path.splitext(str(path_or_buffer))[-1]
                if ext == ".ods":
                    engine = "odf"
        if engine not in self._engines:
            raise ValueError(f"Unknown engine: {engine}")
```

Although presumably that would then throw an exception if the `xlrd` library isn't installed...

requirements.txt
```diff
@@ -9,4 +9,4 @@ pandas==1.1.4
 PyYAML==5.4
 requests==2.25.0
 tqdm==4.52.0
-xlrd==1.2.0
```
thomasdotred, line 12: Can we leave this dependency in just in case some of the other tests (for cocoon) that read excel files?

thomasdotred (changes requested):

requirements.txt
mneedham: Sure.

thomasdotred (approved):

thomasdotred:
waiting for build to pass

PR #232, opened 2021-08-24
Origin/feature/SE-1657 - Make error handling of missing property definitions more explicit

fightingsleep:
# Pull Request Checklist

- [X] Tests pass

# Description of the PR

When checking if a property exists on a given instrument/transaction we need to be more explicit when handling errors. If a 404 is thrown, we will say that the property doesn't exist. If a different status code is returned (ex: Forbidden, Unauthorized, etc), we will continue to throw the exception up the stack.

### DIFF
--- a/tests/unit/cocoon/test_properties.py
+++ b/tests/unit/cocoon/test_properties.py
@@ -6,6 +6,7 @@ import lusidtools.cocoon as cocoon
 import pandas as pd
 import numpy as np
 from lusidtools import logger
+from http import HTTPStatus
 
 
 class CocoonPropertiesTests(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/unit/cocoon/test_properties.py
```diff
@@ -121,16 +131,18 @@ def test_check_property_definitions_exist_in_scope_single(
...
+            (
+                property_existence,
+                property_type,
+            ) = cocoon.properties.check_property_definitions_exist_in_scope_single(
+                api_factory=self.api_factory, property_key=property_key
+            )
+
+            self.assertEqual(property_existence, expected_outcomes[0])
+            self.assertEqual(property_type, expected_outcomes[1])
+        except:
```
MikeMcGarry, line 144: I recommend making this except `ApiException as api_exception` to ensure you are only catching an ApiException and not something you didn't expect.

fightingsleep: This should be fixed now

PR #225, opened 2021-08-06
SE-165: More configurable property_columns

mneedham:
# Pull Request Checklist

- [x] Read the [contributing guidelines](https://github.com/finbourne/lusid-python-tools/blob/master/docs/CONTRIBUTING.md)
- [x] Tests pass

# Description of the PR

Basically lets you pass in property columns like this:

`"column1"`
or
`{"source": "column1", "target": "myAwesomeColumn", "scope": "mySuperScope"}`

You can mix dictionaries and strings in the same list.
If you don’t provide target it will use `source` as property code. If you don’t provide scope it will use `properties_scope` as the scope.

PR #245, opened 2021-11-05
SE-1923 Remove arbitrary request timeout included in err

MikeMcGarry:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

In the merge request to add support for [configurable property columns](https://github.com/finbourne/lusid-python-tools/pull/225) we added an arbitrary 5 second timeout on calls to `SetHoldings` and `AdjustHoldings`.

These were included in error and need to be removed to allow for calls which take 5 seconds or longer to be successful.


jyaffe (changes requested):
Approved in principle but it will need to rebased onto master after [this ](https://github.com/finbourne/lusid-python-tools/pull/246 )is merged so that tests can pass.

jyaffe (approved):

PR #246, opened 2021-11-06
Feature/SENG-37: Update lusid sdk version and fix tests

jyaffe:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Tests were failing in part because an outdated lusid-python-sdk was being used. This change includes an update to the requirements file.

A new model for `transaction_configuration_type_alias` included a new parameter (`source`) which was throwing off a number of tests that expected either the old model or one fewer parameter. 

Also fixed two other tests and added a skip flag to the `properties_search` test that needs further investigation. The functionality being tested has been commented out and a Jira has been raised to investigate.

### DIFF
--- a/lusidtools/lpt/dfq.py
+++ b/lusidtools/lpt/dfq.py
@@ -1,6 +1,7 @@
 import argparse
 import re
 import pandas as pd
+import numpy as np
 
 
 def parse(with_inputs=True, args=None):
@@ -191,6 +192,9 @@ def apply_args(args, given_df):
                         v = int(v[:-7])
                         dflt = 0
                     elif df[col].dtype == int:
+                        v = int(v)
+                        dflt = 0
+                    elif df[col].dtype == np.int64:
                         v = int(v)
                         dflt = 0
                     elif df[col].dtype == float:
--- a/lusidtools/lpt/qry_properties.py
+++ b/lusidtools/lpt/qry_properties.py
@@ -48,4 +48,5 @@ def process_args(api, args):
 
 # Standalone tool
 def main(parse=parse, display_df=lpt.display_df):
-    return lpt.standard_flow(parse, lse.connect, process_args, display_df)
+    raise NotImplementedError("Jira: SENG-40")
+    # return lpt.standard_flow(parse, lse.connect, process_args, display_df)
--- a/lusidtools/lpt/qry_properties.py
+++ b/lusidtools/lpt/qry_properties.py
@@ -5,8 +5,8 @@ from lusidtools.lpt import lpt
 
 # Jira: SENG-40 - Property search was deprecated and this query requires updating
 # Uncomment tooltips when fixed
-#TOOLNAME = "props"
-#TOOLTIP = "Query property definitions"
+# TOOLNAME = "props"
+# TOOLTIP = "Query property definitions"
 
 
 def parse(extend=None, args=None):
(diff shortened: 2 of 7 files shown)


lusidtools/lpt/dfq.py
```diff
@@ -193,6 +194,9 @@ def load_frame(path):
                     elif df[col].dtype == int:
                         v = int(v)
                         dflt = 0
+                    elif df[col].dtype == np.int64:
+                        v = int(v)
+                        dflt = 0
```
thomasdotred, line 199: This is a sensible inclusion, but out of interest, what was it that prompted you to add this?

lusidtools/lpt/qry_properties.py
```diff
@@ -48,4 +48,5 @@ def success(result):
 
 # Standalone tool
 def main(parse=parse, display_df=lpt.display_df):
-    return lpt.standard_flow(parse, lse.connect, process_args, display_df)
+    raise NotImplementedError("Jira: SENG-40")
```
thomasdotred, line 51: If you want we could also comment out TOOLNAME and TOOLTIP. This will prevent the Lusidtools cli from identifying it as a function, removing it from things like the lusidtools --help. Its probably fine to leave as not implemented, but its just another option

lusidtools/lpt/txn_config_yaml.py
```diff
@@ -387,10 +388,11 @@ def alias_con(loader, node):
         return (
             s[0],
             s[1],
+            s[4],
             s[3],
             s[2],
-            unabbrev(s[4]),
-            (len(s) > 5 and s[5] == "default"),
+            unabbrev(s[5]),
+            (len(s) > 6 and s[6] == "default"),
```
thomasdotred, line 395: Would this going to break any existing back-up versions of txn_configs when trying to set configurations? or any jobs that might currently configure TXN types from a YAML file?

lusidtools/lpt/qry_properties.py
thomasdotred: Can we also add a bit more description around this for anyone who doesnt have access to our Jira? something like 
Jira: SENG-40 - Property Search was updated and this is now deprecated and requires updating

tests/integration/commands/test_commands.py
```diff
@@ -143,6 +143,7 @@ def test_get_port_props(self):
         )
         self.validate_results_df(result)
 
+    @unittest.skip("Jira: SENG-40")
```
thomasdotred, line 146: Same here for a bit of description, mainly so we can see in the logs why this gets skipped without having to go back and look at the jira

tests/integration/cocoon/test_upsert_transaction_type_alias.py
```diff
@@ -66,6 +67,7 @@ def setUpClass(cls) -> None:
                         description="TESTSELL1",
                         transaction_class="TESTSELL1",
                         transaction_group="SYSTEM1",
+                        source="SYSTEM1",
```
thomasdotred, line 70: NIce 👍

lusidtools/lpt/dfq.py
jyaffe: This was causing two tests to fail locally - unclear if the test was being run in the pipeline

lusidtools/lpt/txn_config_yaml.py
jyaffe: In theory, yes. In exactly the same way that lpt was broken because of an sdk change, this change could cause issues for lpt users. Good point. Let's discuss

thomasdotred: agreed

PR #258, opened 2022-01-31
SE-2046 Enable filtering on GetTransactions when looking for unresolv…

MikeMcGarry:
…ed instruments

# Pull Request Checklist

- [x] Tests pass

# Description of the PR

When checking to see if any loaded transactions are unresolved, we are currently looking at all transactions in the Portfolio for all time.

We are only interested in whether or not any transactions we just added are unresolved, therefore add a date filter to only look at this subset.

The intent of this is to improve the performance of the call to GetTransactions. 


tests/integration/cocoon/test_cocoon_transactions.py
```diff
@@ -600,10 +601,10 @@ def test_return_unmatched_transactions_extracts_relevant_transactions_and_instru
                 "base_currency": "base_currency",
             },
             file_type="portfolio",
-            mapping_optional={"created": "txn_trade_date"},
+            mapping_optional={"created": "$2000-01-01"},
```
rizwansaeed, line 604: we should probably have a test with and without a date filter with corresponding asserts

MikeMcGarry: Added an additional assertion!

rizwansaeed (approved):

PR #259, opened 2022-02-17
DEV-12265: handle apiurl and bearer token command line arguments

timbarrass:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

lusid-python-tools supports authentication with a bearer token and connection to an apiurl, but doesn't yet enable you to pass those parameters on the command line. This PR enables you to specify them on the command line.


lusidtools/lpt/stdargs.py
```diff
@@ -70,6 +70,13 @@ def __init__(self, description, sections=[]):
             help="choose a special LUSID environment. E.g. 'ipsum_lorem'",
         )
 
+        self.add(
+            "--apiUrl",
+            help="the relevant LUSID Api url (e.g. https://somedomain.lusid.com/api)",
```
steco, line 75: Do you want to say "this is only required it you are supplying a bearer token" (assume that if not, the url comes from the secrets file?)

PR #266, opened 2022-06-10
feature/SE-2251: Update - update instrument upsert to include scope

TobiAdeniyi:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

### DIFF
--- a/lusidtools/cocoon/cocoon.py
+++ b/lusidtools/cocoon/cocoon.py
@@ -1,5 +1,6 @@
 import asyncio
 import concurrent
+from urllib import request
 import uuid
 
 import lusid
@@ -33,7 +34,9 @@ class BatchLoader:
     @staticmethod
     @run_in_executor
     def load_instrument_batch(
-        api_factory: lusid.utilities.ApiClientFactory, instrument_batch: list, **kwargs
+        api_factory: lusid.utilities.ApiClientFactory,
+        instrument_batch: list,
+        **kwargs,
     ) -> lusid.models.UpsertInstrumentsResponse:
         """
         Upserts a batch of instruments to LUSID
@@ -63,7 +66,8 @@ class BatchLoader:
 
         @checkargs
         def get_alphabetically_first_identifier_key(
-            instrument: lusid.models.InstrumentDefinition, unique_identifiers: list
+            instrument: lusid.models.InstrumentDefinition,
+            unique_identifiers: list,
         ):
             """
             Gets the alphabetically first occurring unique identifier on an instrument and use it as the correlation
@@ -92,12 +96,13 @@ class BatchLoader:
             return f"{first_unique_identifier_alphabetically}: {instrument.identifiers[first_unique_identifier_alphabetically].value}"
 
         return api_factory.build(lusid.api.InstrumentsApi).upsert_instruments(
+            scope=kwargs["scope"],
             request_body={
                 get_alphabetically_first_identifier_key(
                     instrument, unique_identifiers
                 ): instrument
                 for instrument in instrument_batch
-            }
+            },
         )
 
     @staticmethod
@@ -506,7 +511,8 @@ class BatchLoader:
             ]
 
             for code, scope in set(
-                [(resource.code, resource.scope) for resource in new_portfolios]
+                [(resource.code, resource.scope)
+                 for resource in new_portfolios]
             ):
 
                 try:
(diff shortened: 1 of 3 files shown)


lusidtools/cocoon/cocoon.py
```diff
@@ -92,12 +96,13 @@ def get_alphabetically_first_identifier_key(
             return f"{first_unique_identifier_alphabetically}: {instrument.identifiers[first_unique_identifier_alphabetically].value}"
 
         return api_factory.build(lusid.api.InstrumentsApi).upsert_instruments(
+            scope=kwargs["scope"],
```
jyaffe, line 99: This seems to be the only change in this file (which is fine). The rest of the changes are formatting ones which might be best to leave because they don't all make sense.

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -742,3 +742,110 @@ def test_load_instrument_lookthrough(self, _, df, mapping):
...
+        )
+
+        # Assert that instruments have been assigne scope
+        self.assertTrue(
+            expr=all(
+                instrument.scope == scope
+                for response in responses["instruments"]["success"]
+                for instrument in response.values.values()
+            )
+        )
```
jyaffe, line 851: Sonar is not enjoying this test because most of it is a duplicate of Lines220-272. 

I think you might be able to just add in L844+ to `test_load_from_data_frame_instruments_success` and you will have what you need.

jyaffe (changes requested):

jyaffe:
You'll need to look at test failures too. Previous tests might load up instruments with a scope provided, under the assumption that the scope would not impact the instrument load. Now that the scope is relevant, it might be that the instruments are not in the default scope where they are expected in the tests.

lusidtools/cocoon/cocoon.py
rizwansaeed: use `.get` in case `scope` is not in the dict

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -742,3 +742,110 @@ def test_load_instrument_lookthrough(self, _, df, mapping):
...
+                {"name": "instrument_name"},
+                {},
+                {"Figi": "figi", "Isin": "isin", "ClientInternal": "client_internal"},
+                ["s&p rating", "moodys_rating", "currency"],
+                "TestPropertiesScope1",
+                expected_response(),
+            ],
+        ]
+    )
+    def test_load_from_data_frame_instruments_scope(
```
rizwansaeed, line 762: add another test case where `scope` is not passed to ensure that works too

lusidtools/cocoon/cocoon.py
```diff
@@ -91,7 +91,9 @@ def get_alphabetically_first_identifier_key(
             first_unique_identifier_alphabetically = unique_identifiers_populated[0]
             return f"{first_unique_identifier_alphabetically}: {instrument.identifiers[first_unique_identifier_alphabetically].value}"
 
+        # If scope is not defined set to default scope
         return api_factory.build(lusid.api.InstrumentsApi).upsert_instruments(
+            scope=kwargs.get("scope") if kwargs.get("scope") != None else "default",
```
rizwansaeed, line 96: I think you can just do 
```
scope=kwargs.get("scope"), "default")
```

as `kwargs` is just a dict

TobiAdeniyi: Screenshot 2022-06-27 at 10 03 53

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -56,19 +56,20 @@ def expected_response(property_scope="TestPropertiesScope1"):
...
 
     @lusid_feature(
-        "T4-1", "T4-2", "T4-3", "T4-4", "T4-5", "T4-6", "T4-7", "T4-8", "T4-9"
+        "T4-1", "T4-2", "T4-3", "T4-4", "T4-5", "T4-6", "T4-7", "T4-8", "T4-9", "T4-10"
     )
     @parameterized.expand(
         [
             [
-                "A standard successful load of instruments",
+                "A standard successful load of scope instruments",
```
jyaffe, line 72: can this say `scoped` instruments?

slemasne:
Hi @TobiAdeniyi 
Looks good - well done!
Perhaps add one more test:

```
Test of upload where instrument scope and properties scope are different
```

jyaffe (approved):

PR #273, opened 2022-07-11
feature/SE-2251: Create - scoped instruments

TobiAdeniyi:
# Pull Request Checklist

- [x]  Tests pass

# Description of the PR


lusidtools/cocoon/cocoon.py
```diff
@@ -1307,6 +1310,7 @@ def load_from_data_frame(
         The api factory to use
     scope : str
         The scope of the resource to load the data into
+        If scope=None and file_type="instrument", scope will be set to the default scope "default"
```
MikeMcGarry, line 1313: Is this actually true?

lusidtools/cocoon/utilities.py
```diff
@@ -76,11 +76,18 @@ def _f(*args, **kwargs):
...
 
+            # Acceptable edge cases
+            # condition 1. check instance matches expected instance type
+            condition_1 = not isinstance(argument_value, argument_details.annotation)
+
+            # condition 2. check annotation is not empty
+            condition_2 = argument_details.annotation is not argument_details.empty
+
+            # condition 3. allow assignment of scope to None if file_type is instrument
+            condition_3 = not (kwargs.get("file_type") in ["Instruments", "instruments"] and argument_name == 'scope')
```
MikeMcGarry, line 87: Do we want to change this behaviour? What is the benefit of adding this? It hasn't been possible to specify this as None up until now, why should we allow it?

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -212,24 +235,39 @@ def test_load_from_data_frame_instruments_success(
...
-            scope=scope,
-            data_frame=data_frame,
-            mapping_required=mapping_required,
-            mapping_optional=mapping_optional,
-            file_type="instruments",
-            identifier_mapping=identifier_mapping,
-            property_columns=property_columns,
-            properties_scope=properties_scope,
-        )
+        if instrument_scope == None:
```
MikeMcGarry, line 246: Better to use `if instrument_scope is None` instead of `==`

tests/integration/cocoon/test_cocoon_instruments.py
```diff
@@ -212,24 +235,39 @@ def test_load_from_data_frame_instruments_success(
...
+                scope=scope,
+                data_frame=data_frame,
+                mapping_required=mapping_required,
+                mapping_optional=mapping_optional,
+                file_type="instruments",
+                identifier_mapping=identifier_mapping,
+                property_columns=property_columns,
+                properties_scope=properties_scope,
+                instrument_scope=instrument_scope,
+            )
```
MikeMcGarry, line 270: Why do we need this condition? If we pass None in as instrument_scope doesn't it default to "default" anyway?

MikeMcGarry (approved):

PR #290, opened 2022-12-06
CTECH-2361: Add a helper method that creates a role calling both the access and identity APIs

df1116:
# Pull Request Checklist

- [ ] Tests pass

# Description of the PR

Currently when creating a role in a python notebook you need to create a role via both the access API and identity API. This has to be done so that the access system (which handles applying policies to a role) can communicate with the identity system (which handles assigned roles to users).

### DIFF
--- a/lusidtools/iam/roles.py
+++ b/lusidtools/iam/roles.py
@@ -1,0 +1,88 @@
+import json
+
+import lusid
+import finbourne_access
+import finbourne_access.utilities
+import finbourne_identity
+
+from finbourne_access import models as access_models
+from finbourne_identity import models as identity_models
+
+from lusidtools.cocoon.utilities import (
+    checkargs,
+)
+import logging
+
+
+@checkargs
+def create_role(
+        api_factory: lusid.utilities.ApiClientFactory,
+        access_role_creation_request: access_models.RoleCreationRequest,
+):
+    """
+
+    Parameters
+    ----------
+    api_factory : lusid.utilities.ApiClientFactory api_factory
+        The api factory to use
+    access_role_creation_request : access_models.RoleCreationRequest
+        The role creation request to use
+
+    Returns
+    -------
+    responses: none
+
+    """
+
+    api_client = api_factory.api_client
+    lusid_api_url = api_client.configuration.host
+    access_token = api_client.configuration.access_token
+
+    # Build the access and identity API URLs from the LUSID API URL.
+    access_api_url = lusid_api_url[: lusid_api_url.rfind("/") + 1] + "access"
+    identity_api_url = lusid_api_url[: lusid_api_url.rfind("/") + 1] + "identity"
+
+    # Build the access and identity API factories.
+    access_api_factory = finbourne_access.utilities.ApiClientFactory(
+        token=access_token,
+        access_url=access_api_url,
+    )
+    identity_api_factory = finbourne_identity.utilities.ApiClientFactory(
+        token=access_token,
+        api_url=identity_api_url,
+    )
+
+    # Build the access and identity roles APIs.
+    access_roles_api = access_api_factory.build(finbourne_access.RolesApi)
+    identity_roles_api = identity_api_factory.build(finbourne_identity.RolesApi)
+
+    access_role_creation_success = False
+    identity_role_creation_success = False
+
+    # Create the role using the access API.
+    try:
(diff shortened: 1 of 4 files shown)


lusidtools/iam/roles.py
```diff
@@ -0,0 +1,88 @@
...
+        identity_role_creation_success = True
+    except finbourne_identity.ApiException as e:
+        detail = json.loads(e.body)
+        if detail["code"] != 157:  # RoleWithCodeAlreadyExists
+            raise e
+        logging.info(
+            f"Role with code {access_role_creation_request.code} has already been created via the identity API"
+        )
+
+    if access_role_creation_success & identity_role_creation_success:
```
adamMQM, line 87: Is this check needed? What happens when an exception is raised?

lusidtools/iam/roles.py
```diff
@@ -0,0 +1,88 @@
...
+        detail = json.loads(e.body)
+        if detail['code'] not in [612, 613, 615]:  # RoleWithCodeAlreadyExists
+            raise e
+        logging.info(
+            f"Role with code {access_role_creation_request.code} has already been created via the access API"
+        )
+
+    # Create the same role using the identity API.
+    identity_role_creation_request = identity_models.CreateRoleRequest(access_role_creation_request.code)
+    try:
```
adamMQM, line 76: Do you need two try-except blocks - is it possible to do this in a single block?

lusidtools/iam/roles.py
```diff
@@ -0,0 +1,88 @@
...
+import logging
+
+
+@checkargs
+def create_role(
+        api_factory: lusid.utilities.ApiClientFactory,
+        access_role_creation_request: access_models.RoleCreationRequest,
+):
+    """
+
```
adamMQM, line 23: Please add a description of what the function does here

https://peps.python.org/pep-0257/#multi-line-docstrings - example

Just because users will get the hint in their IDE which is nice right?

lusidtools/iam/roles.py
```diff
@@ -0,0 +1,88 @@
...
+    Parameters
+    ----------
+    api_factory : lusid.utilities.ApiClientFactory api_factory
+        The api factory to use
+    access_role_creation_request : access_models.RoleCreationRequest
+        The role creation request to use
+
+    Returns
+    -------
+    responses: none
```
adamMQM, line 33: None != none

adamMQM (commented):
Looks good, left a couple of comments for improvement. Also please ensure you run the black formatter on your code.

lusidtools/iam/roles.py
df1116: I have kept it as two different try-except blocks as the exception codes are different for each. It would be possible to combine them, but I figured I made sense to keep them separate especially so that you know obviously which call (access or identity) failed

lusidtools/iam/roles.py
df1116: Check is not needed as such but given that there is no output from the method and that - to answer your next question - on exception we log what went wrong I thought it made sense to log when things went right

tests/integration/iam/test_iam_roles.py
```diff
@@ -0,0 +1,143 @@
...
+
+class IAMTestsRoles(unittest.TestCase):
+    api_factory = None
+    access_api_factory = None
+    identity_api_factory = None
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        secrets_file = Path(__file__).parent.parent.parent.joinpath("secrets.json")
+        print(secrets_file)
```
rizwansaeed, line 25: this looks superfluous

tests/integration/iam/test_iam_roles.py
```diff
@@ -0,0 +1,143 @@
...
+    ):
+        # Build the identity API factory
+        identity_roles_api = self.identity_api_factory.build(finbourne_identity.RolesApi)
+
+        roles = identity_roles_api.list_roles()
+
+        return next((r for r in roles if r.role_id.code == role_code))
+
+
+def create_access_role_creation_request(
```
rizwansaeed, line 126: fix indentation

df1116: Have run `black` on it

rizwansaeed (approved):

PR #296, opened 2023-02-06
CTECH-2727 fix instruments id type 

rizwansaeed:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Fix instrument id type to be str which is failing validation in the SDK

### DIFF
--- a/lusidtools/cocoon/instruments.py
+++ b/lusidtools/cocoon/instruments.py
@@ -83,12 +83,12 @@ def create_identifiers(
     identifiers = {
         # Handles specifying the entire property key e.g. Instrument/default/Figi or just the code e.g. Figi
         prepare_key(identifier_lusid, full_key_format): models.InstrumentIdValue(
-            value=row[identifier_column]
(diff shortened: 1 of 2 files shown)


lusidtools/cocoon/instruments.py
```diff
@@ -83,12 +83,12 @@ def create_identifiers(
...
         # Handles specifying the entire property key e.g. Instrument/default/Figi or just the code e.g. Figi
         prepare_key(identifier_lusid, full_key_format): models.InstrumentIdValue(
-            value=row[identifier_column]
+            value=str(row[identifier_column])
         )
         if file_type == "instrument"
         else row[identifier_column]
         for identifier_lusid, identifier_column in instrument_identifier_mapping.items()
-        if not pd.isna(  # Only use the identifier it it has a value
+        if not pd.isna(  # Only use the identifier it has a value
```
adamMQM, line 91: I think this should probably be `# Only use the identifier if it has a value`

adamMQM (approved):

PR #298, opened 2023-03-17
CTECH-873: update python version to 3.11.2

adamMQM:
# Pull Request Checklist

- [x] Tests pass

# Description of the PR

Updates Dockerfile to Python 3.11.2.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7.0
+FROM python:3.11.2
 
 WORKDIR /usr/src/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,9 +1,7 @@
-FROM python:3.7.0
+FROM python:3.11.2
 
 WORKDIR /usr/src/
 
-RUN apt-get update && apt-get -y install jq
```
pms1969, line 5: I wouldn't get rid of this, there are always security updates to add, and jq is always handy.

pms1969 (approved):

